Issue #7, opened 2017-12-24
Version 0.4


lepture:
With Authlib 0.3, we can create an OAuth 2 server in Flask framework. I'm not going to implement Django integration by version 0.6. In version 0.4, I will focus on OAuth 1 server.

Tasks in version 0.4:

- [x] RFC5849, The OAuth 1.0 Protocol.
- [x] Flask servers
- [x] Test cases
- [x] Documentation

PR #10, opened 2018-01-09
Fix 'multiple values for argument' error in Flask client.

### LINKED ISSUE #9: Unable to pass `fetch_token` and `update_token` to Flask's `register()`
```
  File "/Users/nsantos/.virtualenvs/oauth/lib/python3.6/site-packages/authlib/flask/client/oauth.py", line 77, in create_client
    self.update_token, **kwargs
TypeError: __init__() got multiple values for argument 'fetch_token'
```

nfvs:
Kwarg passed to `register()` takes precedence over the one defined in
the constructor or `init_app()`

Fixes #9

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/flask/client/oauth.py
+++ b/authlib/flask/client/oauth.py
@@ -72,9 +72,12 @@ class OAuth(object):
                 v = self.app.config.get(conf_key, None)
                 kwargs[k] = v
 
+        fetch_token = kwargs.pop('fetch_token', None) or self.fetch_token
+        update_token = kwargs.pop('update_token', None) or self.update_token
+
         client = RemoteApp(
-            name, self.cache, self.fetch_token,
-            self.update_token, **kwargs
+            name, self.cache, fetch_token,
+            update_token, **kwargs
         )
         if compliance_fix:
             client.compliance_fix = compliance_fix


lepture:
No. It is not a bug. The `kwargs` is designed to be the `kwargs` of `RemoteApp` which is the parameters of OAuthClient.

```
:param kwargs: Parameters for :class:`RemoteApp`.

Find parameters from :class:`~authlib.client.OAuthClient`.
```

---

I'd think this is a feature instead. But it is not implemented in the right way. Currently, all clients share the same `fetch_token` and `update_token` methods. If you pass these two methods into `register()`, they should not be shared methods.

https://github.com/lepture/authlib/blob/master/authlib/flask/client/oauth.py#L148

I'll add this feature in [Version 0.4](https://github.com/lepture/authlib/issues/7).

nfvs:
I think you misunderstand the bug. It is a bug because I cannot pass `fetch_token` to `register()`, even if I don't define it in the app.

>  If you pass these two methods into register(), they should not be shared methods.
That's what my PR does. They just get sent to the `RemoteApp` constructor, they are never shared.

nfvs:
```python
self.oauth = OAuth()
self.oauth.init_app(app)
self.microsoft = self.oauth.register(
    'microsoft',
    ....
    fetch_token=self._fetch_token
)
```

This code doesn't work without my PR.

lepture:
@nfvs Yes. Because it is designed that you can not pass `fetch_token` into `register`.

You can check the documentation here: https://docs.authlib.org/en/v0.3/client/frameworks.html#cache-database

Here is my solution, so that you can `register` with `fetch_token` and `update_token`.

https://github.com/lepture/authlib/commit/c14b7eeed598c198d96dcd3eb9da08e532e1d587

Documentation will be available soon.

lepture:
@nfvs With v0.3, your code would be something like this:

```python
self.oauth = OAuth()
self.oauth.init_app(app, fetch_token=self._fetch_token)
self.microsoft = self.oauth.register(
    'microsoft',
    ....
)
```

nfvs:
> Yes. Because it is designed that you can not pass fetch_token into register

Why?

```
register(name, **kwargs)
- kwargs – Parameters for RemoteApp
```
```
class authlib.flask.client.RemoteApp(name, cache=None, fetch_token=None, update_token=None, *args, **kwargs)
```

 `register`'s kwargs are parameters for `RemoteApp`, and `fetch_token` is a parameter of `RemoteApp`, so why isn't that valid?

lepture:
@nfvs 

**First, I've fixed it, so that you can register these two methods now. c14b7ee**

I fixed it. You can pass these parameters now. You can!

---

```
:param kwargs: Parameters for :class:`RemoteApp`.

Find parameters from :class:`~authlib.client.OAuthClient`.
```

> Find parameters from :class:`~authlib.client.OAuthClient`.
> Find parameters from :class:`~authlib.client.OAuthClient`.

nfvs:
Thanks.

PR #72, opened 2018-07-11
Remove client secret check for refresh token

ProstoMaxim:
In case of creating client with authenticate_none method (in password flow), we shouldn't check client's 
secret, because he couldn't have one.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/specs/rfc6749/grants/refresh_token.py
+++ b/authlib/specs/rfc6749/grants/refresh_token.py
@@ -36,9 +36,6 @@ class RefreshTokenGrant(BaseGrant):
         client = self.authenticate_token_endpoint_client()
         log.debug('Validate token request of {!r}'.format(client))
 
-        if not client.has_client_secret():
-            raise UnauthorizedClientError()
-
         if not client.check_grant_type(self.GRANT_TYPE):
             raise UnauthorizedClientError()
 


lepture:
No. You can’t do that. It is insecure.

ProstoMaxim:
I don't get it. I created a client and getting token with password grant, RFC 6749 doesn't require client_secret for this, so I use 'none' token_endpoint_auth_method. Then I want to refresh token for this client and I can't, though 6749 again doesn't required client_secret. Where is discrepancy with RFC?

lepture:
Per https://tools.ietf.org/html/rfc6749#section-6

> The authorization server MUST:
> o  require client authentication for confidential clients or for any
>  client that was issued client credentials (or with other
>   authentication requirements),

lepture:
You can't use a `none` client authentication method with `refresh_token` grant type. In practice, you should only use `none` with `authorization_code` and `implicit` grant types.

lepture:
@ProstoMaxim a public client can't use password/refresh_token flow. You can't change the way it works in Authlib because it is a security issue. If you have to do it in this way, you can subclass the Grant class and rewrite the whole method. But you'd better don't do such thing. Just forget to use a public client to use password, refresh_token grant types.

ProstoMaxim:
Thanks

PR #91, opened 2018-09-26
RFC: Add optional arg to require_oauth

jcassee:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [X] No

---

This PR is a request for comments and does not yet contain tests, etc.

@lepture would you accept some API that validates the access token and makes it available in `current_token` but does not cause an error if no token was found or validation failed.

Open questions:

* An `optional` parameter or another decorator?
* Allow the absence of a token, but do return an error if a provided token was invalid?

### DIFF
--- a/authlib/flask/oauth2/resource_protector.py
+++ b/authlib/flask/oauth2/resource_protector.py
@@ -1,6 +1,6 @@
 import functools
 from contextlib import contextmanager
-from flask import json
+from flask import current_app, json
 from flask import request as _req
 from flask import _app_ctx_stack
 from werkzeug.local import LocalProxy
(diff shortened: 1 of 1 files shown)


lepture:
What is the use case? I can't figure out a use case.

jcassee:
> What is the use case? I can't figure out a use case.

The use case would be an API resource that can be accessed both by anonymous and authenticated users, and that changes its behavior based on that. For example, I have a [HAL](http://stateless.co/hal_specification.html) resources that adds some links to other resources if the request is authenticated to access those resources.

jcassee:
It's weird to see `@require_oauth(required=False)`, but I'll take it. :-) Thanks!

lepture:
I decided to change it to `optional`.

jcassee:
Sounds good, thanks @lepture

PR #101, opened 2019-01-09
More flexibility for session objects

aparamon:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor

- [ ] Yes
- [x] No

---

---

Currently, `OAuth2Session` inherits from [requests.Session](http://docs.python-requests.org/en/master/user/advanced/#session-objects) and uses that session for communicating with both authorization server and resource server. This has the following problems:

1. HTTP clients other than `requests.Session` are not supported, even if they implement the same interface. Consider [Starlette Test Client](https://www.starlette.io/testclient), [aiohttp.ClientSession](https://aiohttp.readthedocs.io/en/stable/client_reference.html#client-session) as examples.

2. Sometimes, an initialized/configured HTTP session is only available as instantiated object. It's impossible to use this from `OAuth2Session` because the latter necessarily creates a new session.

3. Two separate sessions for communicating with authorization server and resource server cannot be used, although this would be highly desired in some work-flows (e.g. unit-testing a resource server).

This PR introduces intermediate class `OAuth2Handler` holding all complicated OAuth 2.0 client logic. Rather than inherit from `requests.Session`, `OAuth2Handler` encapsulates the provided `requests.Session`-like object.

`OAuth2Session` implements both `OAuth2Handler` and `requests.Session`, as before.

New slim `OAuth2HandlerAuth` encapsulates `OAuth2Handler` and implements [requests.auth.AuthBase](http://docs.python-requests.org/en/master/api/#requests.auth.AuthBase) with auto refresh token feature.

`OAuth2Handler` can be used with any `requests.Session`-like objects. `OAuth2HandlerAuth` can be used with any `requests.Session` object, regardless of which HTTP client session is used in its `OAuth2Handler` (separate communication channels for authorization server and resource server).

Please review and comment.


lepture:
This is very interesting. But I have a question, let's take `fetch_access_token` as an example, the last step is

```python
return self._parse_and_validate_token(resp.json())
```

However, `aiohttp.ClientSession` is async, will it really work?

lepture:
I created an example of `OAuth1Protocol` in branch `wip-protocol`. It would be possible to use this `OAuth1Protocol` to create other Session object.

aparamon:
@lepture Thank you for your prompt review!

Indeed, `aiohttp.ClientSession` appears to be bad example because it requires `await` (at least, I fail to figure out how to use it synchronously). But [Starlette Test Client](https://www.starlette.io/testclient) still holds.

I like `OAuth1Protocol` name -- should I rename `OAuth2Handler` -> `OAuth2Protocol`?

For consistency (with `OAuth2Session`), I could also *inherit* `OAuth2HandlerAuth` from `OAuth2Handler`, but I like encapsulation more. Your thoughts?

lepture:
@aparamon It is possible to support async Session with `OAuth1Protocol` since it is sans io. I need to figure out how to implement an aiohttp.ClientSession with Authlib. For now, it is better to use the same way as my work in `wip-protocol`.

aparamon:
@lepture I've looked into `wip-protocol`, and I see you are using different approach: a virtual function `_fetch_token()` to be overridden; while I pass `Session`-like object to be encapsulated. In the former I see the drawback that `_fetch_token()` should be implemented in all descendant classes: both OAuth* `Session` and `AuthBase`.

I'm not sure though how to support `aiohttp.ClientSession` elegantly in any way, though. Probably we shall wait until async operations land onto `requests` ;-)

lepture:
@aparamon I've updated the code in `wip-protocol` branch. It supports `aiohttp` now. Tell me what's your thoughts.

aparamon:
Thanks, I like the way it goes!

`OAuth1Protocol` now takes auth `session` as argument, it's great.

To access protected resources, your method presents another session-like object `OAuth1AsyncClient` (or `OAuth*Client`, generally). However, I found [AuthBase](http://docs.python-requests.org/en/master/api/#requests.auth.AuthBase) to be more flexible; it is the method [recommended by `requests` authors](http://docs.python-requests.org/en/master/user/authentication/#new-forms-of-authentication). Basically, you'll never successfully compete with `requests` in creating better session-like object: better to ally ;-)
So, I propose that `OAuth*ProtocolAuth`s are implemented, on top of corresponding `OAuth*Protocol`s (to gain powerful auto-refresh feature); maybe additionally to `OAuth*Client`s.

Below is a minimal example behind my initial motivation. The important part is that `test_oauth2` arguments may be widely parametrized, allowing `auth_session`, `resource_session` to be very different but compatible implementations of session-like objects.

````py
def test_oauth2(auth_session, resource_session):
    # `auth_session`, `resource_session` are pytest fixtures,
    # and are any request.Session-like objects

    r = resource_session.get('https://resource.company.com/v1/objects')
    assert r.status_code == 401

    oauth2 = OAuth2Protocol(session=auth_session,
                            client=LegacyApplicationClient(client_id='test'))

    # perform OAuth 2.0 dance
    oauth2.fetch_token('https://auth.domain.com/oauth/token',
                       client_id='test', client_secret='grfg', username='user', password='pass',
                       scope=('objects:read'))

    auth = OAuth2ProtocolAuth(oauth2)

    r = resource_session.get('https://server/v1/domains', auth=auth)
    assert r.status_code == 200

    # ...
    # long time passed, token expired...
    # But `auth` conjures auto-refresh spell!
    r = resource_session.get('https://server/v1/domains', auth=auth)
    assert r.status_code == 200
````

Thanks you again for your powerful efforts!

lepture:
This `OAuth1AsyncClient` is a MUST for `aiohttp`. It doesn't share the same API with requests. The `auth` in `aiohttp` can only add `Authorization` to headers.

There is no competition to requests, it is using requests. The problem is that other http libraries don't actually share the same API.

aparamon:
@lepture You are right: unlike in `requests`, in `aiohttp` `auth` is apparently very limited and only supports Basic authorization. I think I'm beginning to grok the need for `Client` abstraction in Authlib; is it correct that `Client` objects will allow to also wrap e.g. [werkzeug.test.Client](http://werkzeug.pocoo.org/docs/0.14/test/#werkzeug.test.Client)?

I have renamed `OAuth2Handler` to `OAuth2Protocol` in order to provide better convergence with your WIP. Let's have `Protocol` objects in the master, so that provides a stable foundation for further ideas/work/PRs (for example OAuth 2.0 `auth` with auto-refresh token feature).

Thank you for your work!

aparamon:
Hi @lepture!
Just took a look at your latest refactorings, and what you have done is awesome!

A tiny proposal: in https://github.com/lepture/authlib/blob/dee7f003b74252e22d62e11a91ec1b629b6fdeae/authlib/oauth2/client_auth.py#L86-L88 doesn't it make sense to init `token` and `token_placement` from `client` if the latter is specified? I.e.:
````py
auth = TokenAuth(client)
````

PR #117, opened 2019-03-14
Fix str + int addition and add option to customize default response_mode via environment variable.

3lpsy:

- [x ] Bugfix
- [ x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [ x] No

---

- [ x] You consent that the copyright of your pull request source code belongs to Authlib's author.

Explanation:
I was experiencing issues with setting the 'exp' claim on the id_token payload where it was trying to add a string and an int. I made a simple change to cast the expiration to an integer. 

In addition, I had a OIDC client that required a response_mode of query. I added an option to set the response_mode via an environment variable. It still defaults to "fragment" if no environment variable is found.

### DIFF
--- a/authlib/oidc/core/grants/util.py
+++ b/authlib/oidc/core/grants/util.py
@@ -1,5 +1,6 @@
 import time
 import random
+import os
 from authlib.oauth2.rfc6749 import InvalidRequestError
 from authlib.oauth2.rfc6749.util import scope_to_list
 from authlib.jose import JWT
@@ -79,7 +80,7 @@ def generate_id_token(
 
 def create_response_mode_response(redirect_uri, params, response_mode=None):
     if response_mode is None:
-        response_mode = 'fragment'
+        response_mode = os.getenv('AUTHLIB_DEFAULT_RESPONSE_CODE', 'fragment')
 
     if response_mode == 'form_post':
         tpl = (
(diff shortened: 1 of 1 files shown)


lepture:
Can you provide the real use case instead? The solution you provided is not good enough, I may figure out a better way.

1. why is your `exp` a string?
2. where did you use `response_mode`?

3lpsy:
I apologize. The "exp" value was a string because I had imported it from env and not cast it to an integer so that's not an issue with authlib (https://github.com/3lpsy/example-openid-server/commit/11367acde35182bd533e22c6ab2607d0dc9e895d). 

For the response_mode, I am using a Apache Guacamole with OpenId Connect (https://guacamole.apache.org/doc/gug/openid-auth.html). When I use the "fragment" response_mode, I fall into an infinite loop. Therefore, I'd like to be able to customize the default response_mode on the authlib side. I believe this is because Guacamole uses Angular and relies heavily on the fragment for front end routing. 

As the current authlib implementation seems to be aligned with the spec, is there a way to specify the response_mode in the server itself where I have something like below?

```
@bp.route('/oauth/authorize', methods=['GET', 'POST'])
def authorize():
    ....
    grant = authorization.create_authorization_response(grant_user=grant_user)
    return grant
```
Regardless, I think it's apparent that the PR is unnecessary and it was an issue with the implementation.

PR #126, opened 2019-05-03
Fixed pseudo RNG being used in token generation

russellcole:
Changed the generate_token function to use a cryptographically strong random number generator from the secrets module, see https://docs.python.org/3.7/library/random.html for more info. Note, this fix requires Python 3.6+.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Other, please describe:
Fixes potential security vulnerability.

- [x] Yes
- [ ] No

The secrets module is new in Python 3.6, since this is a security fix the recommended migration path is to update Python and Authlib.

---

### DIFF
--- a/authlib/common/security.py
+++ b/authlib/common/security.py
@@ -1,12 +1,12 @@
 import os
 import string
-import random
+import secrets
 
 UNICODE_ASCII_CHARACTER_SET = string.ascii_letters + string.digits
 
 
 def generate_token(length=30, chars=UNICODE_ASCII_CHARACTER_SET):
-    rand = random.SystemRandom()
+    rand = secrets.SystemRandom()
     return ''.join(rand.choice(chars) for _ in range(length))
 
 


bjmc:
It's good to be cautious, but I don't think this change is needed. Authlib is already using [`SystemRandom()`](https://docs.python.org/3.7/library/random.html#random.SystemRandom), which relies on [`os.urandom()`](https://docs.python.org/3.7/library/os.html#os.urandom) as its source of randomness.

> This function returns random bytes from an OS-specific randomness source. The returned data should be unpredictable enough for cryptographic applications, though its exact quality depends on the OS implementation.

Authlib is _not_ using the pseudorandom Mersenne Twister implementation provided by `random()`.

Switching to the `secrets` module breaks backwards compatibility, and it doesn't add any additional security over `SystemRandom()` because [it's still using os.urandom() as its entropy source.](https://github.com/python/cpython/blob/3.7/Lib/secrets.py#L47)

russellcole:
That's good to know, I'll close this PR.

PR #127, opened 2019-05-07
Support Custom Grant Type when fetching tokens with Username/Password

lawliet89:
For example, this allows use with some custom grant type with
username/password like Auth0's "Realm Support":
https://auth0.com/docs/api-auth/tutorials/password-grant#realm-support

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Other, please describe:

Allow additional option for users.

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/oauth2/client.py
+++ b/authlib/oauth2/client.py
@@ -157,8 +157,9 @@ class OAuth2Client(object):
         elif username and password:
             if 'scope' not in kwargs and self.scope:
                 kwargs['scope'] = self.scope
+            grant_type = kwargs.pop('grant_type', 'password')
             body = prepare_token_request(
-                'password', body, username=username,
+                grant_type, body, username=username,
                 password=password, **kwargs)
         else:
             grant_type = kwargs.pop('grant_type', 'client_credentials')


lepture:
@lawliet89 This is not a common use case. And the examples on Auth0 are using `grant_type=password`

lawliet89:
I don't think there is any harm allowing the user to override the grant
type.

The code example disagrees with the prose. I need to use their custom grant
type to work with realm.

JonathanHuot:
I don't know if it helps, but in oauthlib, we have moved grant-type which were hard-coded into class variables to help developers to change the `grant_type` values, if they want to. However, I agree it should not be pollute the API/code for these edge-cases.

lawliet89:
Would you mind sharing an example on how to do this?

lawliet89:
@JonathanHuot I disagree that my changes pollutes the code. The way I see it, it's no different from how the PR you linked allows the user to override the `grant_type`.

bjmc:
I think "pollute" might not be a helpful word here, because it has such a negative connotation. [The change]( https://github.com/oauthlib/oauthlib/pull/625/files) in oauthlib allows overriding the `grant_type`, but only by subclassing or monkey-patching the client object. Arguably, that's better because it gives callers who really need it the chance to change that parameter, but it doesn't burden "normal" users with worrying about an extra keyword argument.

On the other hand, an optional keyword with a sensible default value doesn't seem a terrible burden, and the `fetch_token()` method [already accepts a `grant_type` kwarg](https://github.com/lepture/authlib/blob/d3e180477a57faf1b56919d00e2be8574825911f/authlib/oauth2/client.py#L165), but only uses it in cases where `username` and `password` aren't set. As a consumer of the API, I think it would be simpler if overriding `grant_type` were either allowed or disallowed in all cases. As it stands, you have to deal with 'you can override the `grant_type`, but not if you're sending a password', which is a more complex rule.

lawliet89:
> On the other hand, an optional keyword with a sensible default value doesn't seem a terrible burden, and the `fetch_token()` method [already accepts a `grant_type` kwarg](https://github.com/lepture/authlib/blob/d3e180477a57faf1b56919d00e2be8574825911f/authlib/oauth2/client.py#L165), but only uses it in cases where `username` and `password` aren't set. As a consumer of the API, I think it would be simpler if overriding `grant_type` were either allowed or disallowed in all cases. As it stands, you have to deal with 'you can override the `grant_type`, but not if you're sending a password', which is a more complex rule.

Exactly. I had to peruse the source code to see where a preset `grant_type` is coming from. This was rather surprising to me.

JonathanHuot:
@lawliet89, I didn't wanted to add much attention on the pollute word sorry, I just wanted to highlight the key differences between the standards, the customization and their exceptions.

When you read the authlib code, do you prefer to clearly understand the purpose of the standard ? `grant_type=password` is clearly defined and we have a single way to define it.

Or do you prefer to be flooded by yet another-kwargs which has no other values in the specification and thus, leading the developers to be confused on the purpose of the API and how can we use it properly: e.g. do I need to add `grant_type=password` to my function call ? do I have other options ?
 
My belief is that whenever a `kwargs` can be avoided, just avoid it. As @bjmc said, moving this possibility to the class variable to allow subclassing the `OAuth2Client` for a custom use-case seems more reasonnable to me. However that's the @lepture call.

lawliet89:
@JonathanHuot Sure. I agree with you. (Pardon me for my ignorance, what is the relationship between  `authlib` and `oauthlib`?)

In any case, `authlib` currently has the same behaviour for stuff that do not involve username/passwords and for the sake of consistency, i would argue that this PR would make it more consistent and make custom use cases possible.

PR #139, opened 2019-06-29
[security] Fix bug where client can use any valid refresh credential

jonzlin95:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

There is no check that the client & credential.client_id are the same, this adds the check and conforms to [RFC6749](https://tools.ietf.org/html/rfc6749#section-6)

```
The authorization server MUST:
authenticate the client if client authentication is included and
      ensure that the refresh token was issued to the authenticated
      client
```

### DIFF
--- a/authlib/oauth2/rfc6749/grants/refresh_token.py
+++ b/authlib/oauth2/rfc6749/grants/refresh_token.py
@@ -126,7 +126,7 @@ class RefreshTokenGrant(BaseGrant):
             scope = credential.get_scope()
 
         client = self.request.client
-        if client.client_id != credential.client_id:
+        if client and client.client_id != credential.client_id:
             raise InvalidClientError('Invalid client for refresh token')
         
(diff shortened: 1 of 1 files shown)


authlib/oauth2/rfc6749/grants/refresh_token.py
```diff
@@ -125,6 +126,9 @@ def create_token_response(self):
             scope = credential.get_scope()
 
         client = self.request.client
+        if client and client.client_id != credential.client_id:
```
night: should probably be in `_validate_request_token` above, and return a generic invalid grant error to simulate a not found token

lepture:
close it. I've fixed it in https://github.com/lepture/authlib/commit/3a99ab5283c13e6c4b99ea4572743533722544ac

Thanks.

Issue #148, opened 2019-09-14
RFE: integration with FastAPI/Starlette


jonathanunderwood:
[FastAPI ](https://github.com/tiangolo/fastapi)is rapidly gaining popularity as an API framework. It would be great if there was an integration client for FastAPI like there is for Flask etc.

FastAPI doesn't have a plugin system like Flask, but Starlette supports middlewares, and FastAPI supports dependency injection, so I think it should be possible.

lepture:
See #64

lepture:
See #63

jonathanunderwood:
Slightly surprised to see this closed with a reference to #63 - FastAPI bindings donn't require async capability in Authlib (although, if Authlib were async compatible, FastAPI could take advantage of that).

Since you closed this, I assume you're not interested in adding this feature. In which case I'll look into adding them, and submit them as a PR.

Just looking at the code for the Flask bindings, and I have a question: I see the state is stored client side in the session. Is there any reason for doing this rather than storing state server side?

lepture:
@jonathanunderwood Oh, I see. You just need the "client" part, not the server part.

I'm creating a httpx implementation right now. I'll add FastAPI integration in the httpx binding since it supports both sync and async mode.

jonathanunderwood:
Yeah, my use case is for a client binding only, similar to the flask and django ones.

jonathanunderwood:
OK, I have something working for this, not yet PR ready, but I have a question: The flask client has `update_token` support, but the django client doesn't - why is that? I am trying to work out if that's needed for the FastAPI client.

lepture:
@jonathanunderwood Django is using signal to handle token update. Need to add that in documentation.

jonathanunderwood:
Closing this one, since #153 has been merged.

PR #153, opened 2019-10-05
Add Starlette client (#148)

jonathanunderwood:
This commit adds a new Starlette client along with documentation and
tests. The Starlette client is suitable for use with frameworks that
build on top of Starlette such as FastAPI and Bocadillo.

This implements #148.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/integrations/starlette_client/remote_app.py
+++ b/authlib/integrations/starlette_client/remote_app.py
@@ -19,12 +19,6 @@ class RemoteApp(OAuthClient):
 
     """
 
-    def __init__(self, name, fetch_token=None, **kwargs):
-        super(RemoteApp, self).__init__(**kwargs)
-
-        self.name = name
-        self._fetch_token = fetch_token
-
     def save_authorize_state(self, request, redirect_uri=None, state=None):
         """Save ``redirect_uri`` and ``state`` into session during authorization step.
 
--- a/authlib/integrations/starlette_client/remote_app.py
+++ b/authlib/integrations/starlette_client/remote_app.py
@@ -1,64 +1,15 @@
-import functools
 from starlette.responses import RedirectResponse
-from authlib.client import OAuthClient
-from authlib.client.errors import MismatchingStateError
+from authlib.integrations.requests_client import OAuthClient
+from authlib.integrations.client_errors import MismatchingStateError
 
 
-__all__ = ["OAuth", "RemoteApp"]
+__all__ = ["RemoteApp"]
 
 
 _req_token_tpl = "_{}_authlib_req_token_"
 _callback_tpl = "_{}_authlib_callback_"
 _state_tpl = "_{}_authlib_state_"
 _code_verifier_tpl = "_{}_authlib_code_verifier_"
-
-
-class OAuth(object):
-    """Registry for oauth clients.
-
-    Create an instance for registry::
-
-        oauth = OAuth()
-    """
-
-    def __init__(self, fetch_token=None):
-        self._clients = {}
-        self.fetch_token = fetch_token
-
-    def register(self, name, **kwargs):
-        """Register a new remote application.
-
-        :param name: Name of the remote application.
-        :param kwargs: Parameters for :class:`RemoteApp`.
-
-        Find parameters from :class:`~authlib.client.OAuthClient`.
-        When a remote app is registered, it can be accessed with
-        *named* attribute::
-
-            oauth.register('twitter', client_id='', ...)
-            oauth.twitter.get('timeline')
-
-        """
-        client_cls = kwargs.pop("client_cls", RemoteApp)
-        fetch_token = kwargs.pop("fetch_token", None)
-        if not fetch_token and self.fetch_token:
-            fetch_token = functools.partial(self.fetch_token, name)
-
-        compliance_fix = kwargs.pop("compliance_fix", None)
-        client = client_cls(name, fetch_token=fetch_token, **kwargs)
-        if compliance_fix:
-            client.compliance_fix = compliance_fix
-
-        self._clients[name] = client
-        return client
-
-    def __getattr__(self, key):
-        try:
-            return object.__getattribute__(self, key)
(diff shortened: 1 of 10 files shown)


jonathanunderwood:
@lepture it seems you've refactored a lot in the last week during which time I gained work clearance to submit the work on a starlette client. 

It's unclear to me why `_send_token_update` raises a NotImplementedError, though I suspect this is in response to me pointing out the flask client didn't implement it, and I see you've not implemented it for Flask. So, it looks like this is needed for the Starlette client, but it's not really clear to me how that's supposed to work.  Can you explain?

docs/client/starlette.rst
```diff
@@ -0,0 +1,377 @@
+.. _starlette_client:
+
+Starlette OAuth Client
+===================
+
+.. meta::
+    :description: The built-in Starlette integrations for OAuth 1.0 and
+        OAuth 2.0 clients.
+
+.. module:: authlib.starlette.client
```
lepture: here is `authlib.integrations.starlette_client`

lepture:
@jonathanunderwood thanks a lot for your work. I think the most important feature for  Starlette client is **async**. That's why I'm creating a `httpx_client` now. I'll ping you when `httpx_client` is ready. `httpx_client` should contain Async OAuth Client.

authlib/integrations/starlette_client/remote_app.py
```diff
@@ -0,0 +1,110 @@
...
+
+
+class RemoteApp(OAuthClient):
+    """Starlette integrated RemoteApp of :class:`~authlib.client.OAuthClient`.
+
+    This has built-in hooks for ``OAuthClient``.
+
+    """
+
+    def __init__(self, name, fetch_token=None, **kwargs):
```
lepture: from the latest code, you don't need to rewrite this `__init__`.

jonathanunderwood:
> @jonathanunderwood thanks a lot for your work. I think the most important feature for Starlette client is **async**. That's why I'm creating a `httpx_client` now. I'll ping you when `httpx_client` is ready. `httpx_client` should contain Async OAuth Client.

I agree that would be nice to have. Could we merge this now and migrate the integration  to the new `httpx_client` client when it is available? - this is already usable with the Starlette framework family, even if it is synchronous.

docs/client/starlette.rst
jonathanunderwood: Fixed.

lepture:
@jonathanunderwood yes, I can merge it for now. But there won't be a new release soon. I hope I can make a new release in next month. Are you going to use the master code in your project?

jonathanunderwood:
> @jonathanunderwood yes, I can merge it for now. But there won't be a new release soon. I hope I can make a new release in next month. Are you going to use the master code in your project?

That's great to hear - it means I won't have to redo the legal review at work at a later date for this specific code. Yes, I'll be using the master branch for our project.

lepture:
@jonathanunderwood everything looks good for me. But if you could provide a more "starlette-like" database documentation, that would be great. The current documentation is using Django's database design. For example, Flask is using SQLAlchemy for the database example.

jonathanunderwood:
I've made the database design section less like the Django one. But, it's to be noted that Starlette has now ORM support at all, and so is agnostic. I am not convinced that section of the documentation really has much value, but it'll be a starting point to improve on.

In all of the client docs, actually, the database sections don't really make a lot of sense. For example, this part:

```
        def to_token(self):
            return dict(
                oauth_token=self.access_token,
                oauth_token_secret=self.alt_token,
            )
```
which refers to attributes that don't exist for that class. This is true for flask and django client docs. If I'm honest, the documentation for the clients could be more helpful if restructured and significantly re-written, but that's out of scope for this PR, I think.

lepture:
@jonathanunderwood I merged it for now.

Do you have any idea on how to restructure the documentation?

jonathanunderwood:
> Do you have any idea on how to restructure the documentation?

Well, the problem with the set of client docs at the moment is that there's a lot of cut+paste between them, with small tweaks. That indicates that it might be better to have a single document for all the clients that communicates the common concepts across the clients, and then some framework specific notes for each framework.

The other big problem with the docs at present is that they are sketches of how the library could be used, rather than directly run-able code. It would be better to have some super simple but complete code for each framework demonstrating how the clients should be used.

The FastAPI documentation is a great example of directly useable documentation.

jonathanunderwood:
Also: many thanks for merging, and your feedback on all of this. And your work on this library.

lepture:
> The other big problem with the docs at present is that they are sketches of how the library could be used, rather than directly run-able code.

Yes, because it is not a tutorial. For now, there is no real tutorial for Authlib. It is good to have a tutorial for each part of Authlib. But tutorial takes much time to write.

I'm still working on this project in my part time. Although I've gotten some customers via https://authlib.org/plans now, it is not enough to support my life right now. 

I will rewrite the documentation again when v0.13 released. With v0.13, there should be only 2 or 3 RFCs left. v0.13 is a huge release, it is the very step towards v1.0.

lepture:
@jonathanunderwood 

> the problem with the set of client docs at the moment is that there's a lot of cut+paste between them, with small tweaks.

I have updated the documentation for web framework integrations:

1. https://docs.authlib.org/en/latest/client/frameworks.html
2. https://docs.authlib.org/en/latest/client/flask.html
3. https://docs.authlib.org/en/latest/client/django.html

Now, Flask and Django share the same documentation now. Does this make sense to you?

Starlette documentation is not updated yet. I'll change the documentation when it becomes "async".

For ready to use examples, I'm going to create example repositories for every framework.

jonathanunderwood:
That sounds great, I'll read over them when I get a chance

lepture:
@jonathanunderwood here is the re-implementation of the starlette client https://github.com/lepture/authlib/pull/155

PR #186, opened 2020-01-30
Starlette: Force GET on authorize_redirect

phy25:
Starlette uses 307 status code for RedirectResponse by default:
https://www.starlette.io/responses/#redirectresponse

This will reuse the same method of the original request:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/307

...while RFC6749 only ensures that GET is accepted (POST is optional):
https://tools.ietf.org/html/rfc6749#section-3.1

By using HTTP_303_SEE_OTHER the redirect will always occur using a GET:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/303

> on README.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/integrations/starlette_client/remote_app.py
+++ b/authlib/integrations/starlette_client/remote_app.py
@@ -1,4 +1,5 @@
 from starlette.responses import RedirectResponse
+from starlette.status import HTTP_303_SEE_OTHER
 from ..asgi_client import AsyncBaseApp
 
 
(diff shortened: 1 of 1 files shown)


phy25:
I am not sure why CI check is complaining:

> Please provide an upload token from codecov.io with valid arguments

lepture:
Fixed in v0.14

PR #192, opened 2020-02-12
Add sync HTTPX AssertionClient and OAuth2Client

ulodciv:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No (I think)

---

### DIFF
--- a/authlib/integrations/httpx_client/__init__.py
+++ b/authlib/integrations/httpx_client/__init__.py
@@ -9,9 +9,9 @@ from authlib.oauth1 import (
 from .oauth1_client import OAuth1Auth, AsyncOAuth1Client
 from .oauth2_client import (
     OAuth2Auth, OAuth2ClientAuth,
-    AsyncOAuth2Client,
+    AsyncOAuth2Client, OAuth2Client,
 )
-from .assertion_client import AsyncAssertionClient
+from .assertion_client import AsyncAssertionClient, AssertionClient
 from ..base_client import OAuthError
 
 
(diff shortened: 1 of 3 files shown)


phy25:
I think the sync support was dropped intentionally: https://github.com/lepture/authlib/releases/tag/v0.14

ulodciv:
> I think the sync support was dropped intentionally: https://github.com/lepture/authlib/releases/tag/v0.14

Yea I saw that, but why? HTTPX is meant to support sync and async as of [0.11](https://github.com/encode/httpx/blob/master/CHANGELOG.md#0110-january-9th-2020) ("HTTPX is a fully featured HTTP client for Python 3, which provides sync and async APIs, and support for both HTTP/1.1 and HTTP/2."). My understanding is that this will stay as it is necessary for HTTPX to remain a quasi drop-in replacement for requests (see [the first item](https://github.com/encode/httpx#features) in the feature list).

lepture:
HTTPX changes very quickly, it once dropped support for sync mode, that is when Authlib removed sync client. I'm not sure if they will drop sync client again. I'll wait until it reaches 1.0.

ulodciv:
@lepture 
 
I totally understand what you mean, but the fact remains that the main path to adoption of HTTPX from requests is to use HTTPX's non-async objects, I think. In any case, I just put the additional objects proposed in this PR in my projects for now.

Issue #208, opened 2020-03-20
incompatibility with new httpx version


suud:
**Describe the bug**

`httpx`'s public api has been updated (see https://github.com/encode/httpx/issues/772). `httpx.content_streams` is a private module, since Version `0.12.1` of `httpx`.

I'm wondering what a reliable fix would look like. Importing (from) the private module `httpx._content_streams` might break in the future.

Here, some more cases where we use `httpx` modules that are now private (there might be more):
- https://github.com/lepture/authlib/blob/f754eff0c21a11c837a25bc6d4a49875de243c72/authlib/integrations/httpx_client/oauth2_client.py#L3-L6
- https://github.com/lepture/authlib/blob/142a7e5b8d4290707808baa5b35635961bcb279a/tests/py3/utils.py#L3
**Error Stacks**

```python
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/usr/local/lib/python3.8/site-packages/authlib/integrations/httpx_client/__init__.py", line 9, in <module>
    from .oauth1_client import OAuth1Auth, AsyncOAuth1Client
  File "/usr/local/lib/python3.8/site-packages/authlib/integrations/httpx_client/oauth1_client.py", line 11, in <module>
    from .utils import extract_client_kwargs, rebuild_request
  File "/usr/local/lib/python3.8/site-packages/authlib/integrations/httpx_client/utils.py", line 2, in <module>
    from httpx.content_streams import ByteStream
ModuleNotFoundError: No module named 'httpx.content_streams'
```

**To Reproduce**

```python
from authlib.integrations.httpx_client import AsyncOAuth2Client
```

**Environment:**

 - OS: Debian 10 (Buster)
 - Python Version: 3.8
 - Authlib Version: 0.14.1
 - httpx Version: 0.12.1

lig:
Could somebody teach `httpx` authors that this change should've bumped version to 0.13.0, please?

phy25:
Not sure why @lig doesn't want to reach out, but:

> Note: HTTPX should be considered in beta. We believe we've got the public API to a stable point now, but would strongly recommend pinning your dependencies to the 0.12.* release, so that you're able to properly review API changes between package updates. A 1.0 release is expected to be issued sometime on or before April 2020.

https://github.com/encode/httpx/blob/master/README.md

lig:
> pinning your dependencies to the 0.12.*

the fun is that 0.12.0 and 0.12.1 are incompatible

[...]

PR #209, opened 2020-03-20
Import private httpx modules

suud:
The public api of httpx has changed in version 0.12.1.

Fixes: https://github.com/lepture/authlib/issues/208

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [x] Yes
- [ ] No

- ~~Not compatible with `httpx<0.12.0`: the `httpx` library has to be updated to a version `>=0.12.0` in existing applications.~~ **EDIT:** Fixed with https://github.com/lepture/authlib/pull/209/commits/116988b61ab8f0910575088c31937c28d84fcc9d as suggested by @phy25 in https://github.com/lepture/authlib/pull/209#issuecomment-601755350
- Depends on some private modules from `httpx`: as explained in https://github.com/lepture/authlib/issues/208. I hope somebody will come up with a better solution than mine!

---

### DIFF
--- a/authlib/integrations/httpx_client/utils.py
+++ b/authlib/integrations/httpx_client/utils.py
@@ -1,5 +1,5 @@
 from httpx import URL
-from httpx.content_streams import ByteStream
+from httpx._content_streams import ByteStream
 from authlib.common.encoding import to_bytes
 
 
--- a/authlib/integrations/httpx_client/utils.py
+++ b/authlib/integrations/httpx_client/utils.py
@@ -1,5 +1,8 @@
 from httpx import URL
-from httpx._content_streams import ByteStream
+try:
+    from httpx._content_streams import ByteStream
+except ImportError:
+    from httpx.content_streams import ByteStream
 from authlib.common.encoding import to_bytes
 
 
--- a/tests/py3/utils.py
+++ b/tests/py3/utils.py
@@ -1,6 +1,6 @@
 import json
 from httpx import Response
-from httpx.dispatch.base import AsyncDispatcher
+from httpx._dispatch.base import AsyncDispatcher
 
 
 class MockDispatch(AsyncDispatcher):
--- a/tests/py3/utils.py
+++ b/tests/py3/utils.py
@@ -1,6 +1,9 @@
 import json
 from httpx import Response
-from httpx._dispatch.base import AsyncDispatcher
+try:
+    from httpx._dispatch.base import AsyncDispatcher
+except ImportError:
+    from httpx.dispatch.base import AsyncDispatcher
 
 
 class MockDispatch(AsyncDispatcher):
--- a/authlib/integrations/httpx_client/oauth2_client.py
+++ b/authlib/integrations/httpx_client/oauth2_client.py
@@ -1,10 +1,9 @@
 import typing
-from httpx import (
-    AsyncClient,
-    Auth,
-    Request,
-    Response,
-)
+from httpx import AsyncClient, Auth
+try:
+    from httpx import Request, Response
+except ImportError:
+    from httpx.models import Request, Response
 from authlib.common.urls import url_decode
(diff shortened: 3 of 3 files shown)


phy25:
I think a `try...catch ImportError...` with attempts of different namespaces would be better?

pdiwan:
@lepture could you please get this fix merged? This is a blocking bug for me.

lepture:
@pdiwan you can use the old version of httpx. It shouldn't be a blocking issue.

pdiwan:
I could but just trying to avoid hardcoding versions now and later change when the fix is merged.

authlib/integrations/httpx_client/utils.py
```diff
@@ -1,5 +1,8 @@
 from httpx import URL
-from httpx.content_streams import ByteStream
+try:
+    from httpx._content_streams import ByteStream
```
lepture: I don't think it is a good idea to import from private modules. I need to ask @encode team to make it public.

lepture:
Please pin httpx version.

tests/py3/utils.py
```diff
@@ -1,6 +1,9 @@
 import json
 from httpx import Response
-from httpx.dispatch.base import AsyncDispatcher
+try:
+    from httpx._dispatch.base import AsyncDispatcher
+except ImportError:
+    from httpx.dispatch.base import AsyncDispatcher
```
lepture, line 6: This module will be changed again in v0.13. That's why I'd like to keep our current implementation.

authlib/integrations/httpx_client/utils.py
tomchristie: So, it's not obvious to me that the `rebuild_request` function needs to be written in a way that uses private API here. Can you change that function instead? Perhaps like so...

```
def rebuild_request(request, url, headers, body):
    method = request.method
    headers = Headers(request.headers)
    headers.update(headers)
    return Request(method=method, url=url, headers=headers, body=body)
```

tests/py3/utils.py
tomchristie: An alternative to using the private dispatch API here would be to implement the stubs as either ASGI or WSGI applications, and use `Client(app=wsgi_app)` or `AsyncClient(app=asgi_app)`.

authlib/integrations/httpx_client/utils.py
lepture: @tomchristie unfortunately, there is no `body` parameter in `Request`.

tomchristie: Let me rephrase that. `content=...`

lepture: @tomchristie 

```
class Request:
    def __init__(
        self,
        method: str,
        url: typing.Union[str, URL],
        *,
        params: QueryParamTypes = None,
        headers: HeaderTypes = None,
        cookies: CookieTypes = None,
        data: RequestData = None,
        files: RequestFiles = None,
        json: typing.Any = None,
        stream: ContentStream = None,
    ):
```

This is the parameters of `Request`, there is no suitable parameter for `body`.

lepture: @tomchristie there is no `content` either, the only available parameter would be `data=` (I think). But this `rebuild_request` is used in `auth_flow`, and it is `requires_request_body = True`. The `request` returned by `rebuild_request` is not `read` or `aread` yet, thus this `request` would not have `.content` property.

lepture: @tomchristie it would also be helpful if we can add a property setter on `Request`:

```py
    @content.setter
    def content(self, value: bytes) -> None:
        self._content = value
        self.stream = ByteStream(self._content)
```

Do you think it is ok to add this setter?

PR #214, opened 2020-03-29
Fix #213: Fixing auto-refresh token for client-credentials.

### LINKED ISSUE #213: ensure_active_token doesn't work for client_credentials
**Describe the bug**

Currently, authlib is unable to fetch a new access token if the previous one expires when the client is using a `client_credentials` `grant_type`, because at the point it tries to fetch it, it doesn't pass along the `client_id` and `client_secret` with the request to fetch it - the `kwargs` argument is empty at this point. This leads to the Authentication Server responding with an `invalid_token` error.

**Error Stacks**

Instead of an error stack, which is not too relevant, I'll point out to the exact place where the bug is:

1. Here is where the `ensure_active_token` method uses the kwargs to fetch the token: https://github.com/lepture/authlib/blob/master/authlib/integrations/requests_client/oauth2_session.py#L28
1. And here is where the class calls that method without passing along any kwargs, which leads to the call to `fetch_token` not have any `client_id` or `client_secret`, which are necessary for `OAuth2Client._prepare_token_endpoint_body`: https://github.com/lepture/authlib/blob/master/authlib/integrations/requests_client/oauth2_session.py#L35

**To Reproduce**

To reproduce the error, first you have to have a way to expire the token that you get from the auth server. In my case, I store the token in an encrypted Redis item, so I just change it to expire in the past, but you could just wait for the token to expire too - or, if you have access to the auth server running locally, just make it expire tokens in a few seconds or so.

Then, to reproduce:
1. Make sure you have your `client_kwargs` configuration contains a `client_credentials` `grant_type`;
1. Make sure the RemoteApp has access to the access token;
1. Make a request to the data server;
1. Ensure the access token expires somehow (see above);
1. Make another request to the data server

**Expected behavior**

I expect the library to be able to fetch a new access token correctly.

**Environment:**

 - OS: Arch Linux 5.5.13-arch1-1 #1 SMP PREEMPT Wed, 25 Mar 2020 16:04:40 +0000 x86_64 GNU/Linux
 - Python Version: 3.6.10
 - Authlib Version: 0.14.1

**Additional context**

I would try to send a PR myself, but I don't know how to approach this problem yet; The thing is, it seems that it's the RemoteApp that has access to client ID and secret, but the attempt to re-fetch a token is made inside the OAuth2Client itself, so it seems that it doesn't have access to that information.

[EDIT] Just pushed a PR with the fix. I tested it locally, and it works, it fixes the issue.

diogobaeder:
This change ensures that the client_id and client_secret kwargs are passed along
to the request for fetching a new access token when the previous one has expired.

> on README.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/integrations/requests_client/oauth2_session.py
+++ b/authlib/integrations/requests_client/oauth2_session.py
@@ -24,6 +24,8 @@ class OAuth2Auth(AuthBase, TokenAuth):
                 client.refresh_token(url, refresh_token=refresh_token, **kwargs)
             elif client.metadata.get('grant_type') == 'client_credentials':
                 access_token = self.token['access_token']
+                kwargs.setdefault('client_id', self.client.client_id)
+                kwargs.setdefault('client_secret', self.client.client_secret)
                 token = client.fetch_token(
                     url, grant_type='client_credentials', **kwargs)
                 if client.update_token:


diogobaeder:
Hi @lepture , sorry to nag you on this, but can you review and possibly merge this if it looks good for you? It's a very important change for a project I work on...

diogobaeder:
I don't see `audience` being part of the standard, so I don't think it would be a good idea to use it directly; However, being able to pass along arbitrary kwargs sounds like a good idea to me. To be honest, I'm not even sure why it's not passing along `client_kwargs`, as it should, IIUC.

lepture:
I don't think it is a good fix. I'll work on to it soon.

diogobaeder:
So can you help us with the fix? I'm OK with whatever solution fixes the underlying bug.

lepture:
Current code itself has no problem. It is caused by your usage. It seems that the third party website doesn't accept `client_secret_basic` auth method, it only accept `client_secret_post` method. In this case you can create a `session` with:

```py
session = OAuth2Session(
    # .....
    token_endpoint_auth_method="client_secret_post",
)
```

diogobaeder:
I really wanted to test whether or not that configuration actually fixes the issue I had, but I just couldn't manage to reproduce that issue again. So I guess I'll have to blindly accept the resolution then.
Thanks anyway!

PR #227, opened 2020-05-03
🐛 fix for "expires_at" / "expires_in" param when "None" is specified

stefanitsky:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Description**
Some providers provide endless access tokens, therefore, if you specify None for expires_at or expires_in - a token initialization error occurs.

For example, such services are GitHub and Gitlab. Moreover, in the case of the Gitlab, the token for updating is returned, but without expires_at (long-live bug and it won't be fixed, probably).

**Links**:
* Gitlab token: https://stackoverflow.com/questions/39264452/gitlab-oauth-access-token-validity
* GitHub token: https://stackoverflow.com/questions/26902600/whats-the-lifetime-of-github-oauth-api-access-token

### DIFF
--- a/authlib/oauth2/rfc6749/wrappers.py
+++ b/authlib/oauth2/rfc6749/wrappers.py
@@ -5,9 +5,9 @@ from .errors import InsecureTransportError
 
 class OAuth2Token(dict):
     def __init__(self, params):
-        if 'expires_at' in params:
+        if params.get('expires_at'):
             params['expires_at'] = int(params['expires_at'])
-        elif 'expires_in' in params:
+        elif params.get('expires_in'):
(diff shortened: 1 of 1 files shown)


lepture:
Thanks

fboudra:
Does it mean we can implement like [Gitlab/Doorkeeper](https://github.com/doorkeeper-gem/doorkeeper/wiki/Customizing-Token-Expiration#access-token) is doing and use [authlib/OAuth 2 server](https://docs.authlib.org/en/latest/flask/2/authorization-server.html#token) to provide a custom access token expiration (endless)?

PR #228, opened 2020-05-05
Enable an option to config the leeway for claim validation

1wpro2:
## Issues

Fixes: #229 

## Background

At the moment, the existing valid method allows a passed in configuration to set some leeways.

For example:
https://github.com/lepture/authlib/blob/36e2236b56274879688ecb8019bf1e9bc8e43746/authlib/oidc/core/claims.py#L26

However, this has been hardcoded to `2 minutes` in the invocation.

For example:
https://github.com/lepture/authlib/blob/36e2236b56274879688ecb8019bf1e9bc8e43746/authlib/integrations/base_client/remote_app.py#L200

## Changes

This will enable an option for developers to configure the leeway as designated.

So that when we call the `parse_id_token`, in addition to the existing option
```
        try:
           # this has a default leeway of 2 minutes
            userinfo = oauth.keycloak.parse_id_token(token) 
        except Exception as e:
            log.error(f"Got a malformed response", e)
```

we can also set up

```
        try:
           # this has a configured leeway of 0 minutes
            userinfo = oauth.keycloak.parse_id_token(token, leeway=0) 
        except Exception as e:
            log.error(f"Got a malformed response", e)
```

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---


authlib/integrations/base_client/remote_app.py
```diff
@@ -197,5 +197,5 @@ def load_key(header, payload):
             claims_options=claims_options,
             claims_params=claims_params,
         )
-        claims.validate(leeway=120)
+        claims.validate(leeway)
```
lepture, line 200:
```suggestion
        claims.validate(leeway=leeway)
```

Because the first value of `validate` is `now`.

lepture:
eh, tests passed. I need to improve the test case, your change should fail the test case.

authlib/integrations/base_client/remote_app.py
1wpro2: Good catch. Thanks.

PR #270, opened 2020-09-10
#268 #269 Improve HTTPX Support

berzoidberg:
Fixes #268
Fixes #269  

**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

- [x] You consent that the copyright of your pull request source code belongs to Authlib's author.

PR #273, opened 2020-09-16
Fix test if authorization should be provided to the underlying httpx client

asyncee:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [x] Yes, for users of very old HTTPX version (0.11.x)
- [ ] No

In HTTPX 0.11.x `auth` argument to the `AsyncClient.request` had `None` as default. Since 0.12 default is changed to a sentinel object of type `UnsetType`. Users should upgrade their HTTPX version. 

We are using HTTPX 0.14 and faced the issue when authlib's `AsyncOauth2Client` had a token but never provided it to the underlying client because `auth` argument to `AsyncOauth2Client.request` method always had a value of `httpx._config.UNSET` and not `None` (code expects it to be `None`). 

HTTPX version 0.14.3
Authlib version 0.14.3

### DIFF
--- a/authlib/integrations/httpx_client/oauth2_client.py
+++ b/authlib/integrations/httpx_client/oauth2_client.py
@@ -1,6 +1,6 @@
 import asyncio
 import typing
-from httpx import AsyncClient, Auth, Request, Response
+from httpx import AsyncClient, Auth, Request, Response, _config
 from authlib.common.urls import url_decode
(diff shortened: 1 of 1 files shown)


lepture:
It is already done by https://github.com/lepture/authlib/pull/270. Thanks.

PR #279, opened 2020-10-11
Improve error reporting in tests

V02460:
Some tests throw unhelpful error messages when an expected exception is not thrown. This PR makes these tests report the correct error. It changes the use of the `self.assertRaises` context manager to be aligned with its [documented usage](https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertRaises) (namely, using `cm` outside of the `with`-statement).

**Previously:**
`AttributeError: '_AssertRaisesContext' object has no attribute 'exception'`
**Now:**
`AssertionError: ValueError not raised`

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---


lepture:
Thanks

PR #287, opened 2020-10-21
Add missing methods to HTTPX sync clients

ulodciv:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

---

### DIFF
--- a/authlib/integrations/httpx_client/assertion_client.py
+++ b/authlib/integrations/httpx_client/assertion_client.py
@@ -87,3 +87,14 @@ class AssertionClient(_AssertionClient, Client):
             auth = self.token_auth
         return super(AssertionClient, self).request(
             method, url, auth=auth, **kwargs)
+    
+    def _refresh_token(self, data):
+        resp = self.request("POST", self.token_endpoint, data=data, withhold_token=True)
+
+        token = resp.json()
+        if "error" in token:
+            raise OAuth2Error(
+                error=token["error"], description=token.get("error_description")
+            )
+        self.token = token
+        return self.token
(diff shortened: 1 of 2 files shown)


lepture:
We don't need them, they are already in the parent class.

ulodciv:
@lepture I hear you, but these methods are present in the async clients, and I see no obvious reason why they would be absent from the sync clients. Furthermore, they don't work in my code as is, whereas they work with my version of them, which is exactly the same, except for the additional methods I have. 

I had them included in the the PR https://github.com/lepture/authlib/pull/192 I've been using them, and still have to use them since the ones that were added recently don't work.

authlib/integrations/httpx_client/assertion_client.py
```diff
@@ -87,3 +87,14 @@ def request(self, method, url, withhold_token=False, auth=None, **kwargs):
             auth = self.token_auth
         return super(AssertionClient, self).request(
             method, url, auth=auth, **kwargs)
+    
+    def _refresh_token(self, data):
```
lepture, line 91: see: https://github.com/lepture/authlib/blob/master/authlib/oauth2/rfc7521/client.py#L72

authlib/integrations/httpx_client/oauth2_client.py
```diff
@@ -212,3 +212,47 @@ def ensure_active_token(self):
                 self.update_token(token, access_token=access_token)
         else:
             raise InvalidTokenError()
+
+    def _fetch_token(
```
lepture, line 216: see: https://github.com/lepture/authlib/blob/master/authlib/oauth2/client.py#L207

lepture:
> they don't work in my code as is

Is it possible to create a test case in the `tests` folder?

Issue #309, opened 2021-01-10
oidc raises exception when auth_time claim is decimal


tntclaus:
**Describe the bug**

According to [spec](https://openid.net/specs/openid-connect-core-1_0.html#IDToken) 'auth_time' is JSON NUMBER which can be decimal. While most oidc providers set it to `int`, there are some (e.g. JetBrains Hub) that provide it in decimal.

**Error Stacks**

```
raceback (most recent call last):
  File "/usr/local/lib/python3.8/site-packages/synapse/handlers/oidc_handler.py", line 671, in handle_oidc_callback
    userinfo = await self._parse_id_token(token, nonce=nonce)
  File "/usr/local/lib/python3.8/site-packages/synapse/handlers/oidc_handler.py", line 474, in _parse_id_token
    claims.validate(leeway=120)  # allows 2 min of clock skew
  File "/usr/local/lib/python3.8/site-packages/authlib/oidc/core/claims.py", line 41, in validate
    self.validate_auth_time()
  File "/usr/local/lib/python3.8/site-packages/authlib/oidc/core/claims.py", line 66, in validate_auth_time
    raise InvalidClaimError('auth_time')
authlib.jose.errors.InvalidClaimError: invalid_claim: Invalid claim "auth_time"
```

**To Reproduce**

id_token:

```
{'sub': 'redacted-aff5-461c-8f6c-4e143fad41df', 'name': 'Klaus Schwartz', 'preferred_username': 'klaus', 'profile': 'https://example.com/hub/users/redacted-aff5-461c-8f6c-4e143fad41df', 'picture': 'data:image/jpeg;base64,image-data', 'email': 'redacted@example.com', 'email_verified': True, 'iss': 'https://example.com/hub', 'aud': ['bb4872fc-12a6-4f74-8744-96263d20807b'], 'exp': 1617590286.706, 'iat': 1610302441.104, 'auth_time': 1609814286.706, 'nonce': 'redacted'}
```

**Expected behavior**

`auth_time` validation raises no errors

**Environment:**

 - OS: any
 - Python Version: 3.8
 - Authlib Version: master

**Additional context**

JetBrains Hub sets `auth_time` as seconds with millis in decimal part

PR #310, opened 2021-01-10
Fix claims.py to support decimal values of `auth_time` claim

tntclaus:
according to spec 'auth_time' is JSON NUMBER which can be decimal

fixe #309

> on README.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/oidc/core/claims.py
+++ b/authlib/oidc/core/claims.py
@@ -56,7 +56,7 @@ class IDToken(JWTClaims):
         if self.params.get('max_age') and not auth_time:
             raise MissingClaimError('auth_time')
 
-        if auth_time and not (isinstance(auth_time, int) or isinstance(auth_time, float)):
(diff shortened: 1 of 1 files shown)


authlib/oidc/core/claims.py
```diff
@@ -56,7 +56,7 @@ def validate_auth_time(self):
         if self.params.get('max_age') and not auth_time:
             raise MissingClaimError('auth_time')
 
-        if auth_time and not isinstance(auth_time, int):
+        if auth_time and not (isinstance(auth_time, int) or isinstance(auth_time, float)):
```
lepture, line 59: `isinstance(auth_time, (int, float))`

tntclaus: sorry, I have few to no experience with python. applied this simplification.

PR #325, opened 2021-02-26
Use resp.raise_for_status() before using resp.json().

pope1ni:
It is better to raise an exception due to an HTTP error status rather
than an error like the following which may be received due to a 502 Bad
Gateway which returns HTML content, for example.

json.decoder.JSONDecodeError: Expecting value: line 2 column 1 (char 1)

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---


lepture:
Thanks

Issue #391, opened 2021-09-28
Allow passing in additional header to the JWT client assertion


janwijbrand:
I'd like to pass in addtional JWT headers into the client assertion, such as the `kid` so the recipient of the token can find the corresponding key at the registered JWKS endpoint.

**Describe the solution you'd like**

When setting up an `PrivateKeyJWT()` instance to register as client auth method, I'd like to pass in additional JWT headers such as the `kid`.

I'll reference a PR that implements this against `maintain-0.15`.

PR #392, opened 2021-09-28
Allow passing in additionl jwt headers

janwijbrand:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

Referencing https://github.com/lepture/authlib/issues/391

I'd like to pass in addtional JWT headers into the client assertion, such as the kid so the recipient of the token can find the corresponding key at the registered JWKS endpoint.

Describe the solution you'd like

When setting up an PrivateKeyJWT() instance to register as client auth method, I'd like to pass in additional JWT headers such as the kid.

### DIFF
--- a/authlib/oauth2/rfc7523/assertion.py
+++ b/authlib/oauth2/rfc7523/assertion.py
@@ -5,12 +5,11 @@ from authlib.common.security import generate_token
 
 def sign_jwt_bearer_assertion(
         key, issuer, audience, subject=None, issued_at=None,
-        expires_at=None, claims=None, header=None, **kwargs):
+        expires_at=None, alg=None, header=None, claims=None, **kwargs):
 
     if header is None:
         header = {}
-    alg = kwargs.pop('alg', None)
-    if alg:
+    if alg is not None:
         header['alg'] = alg
     if 'alg' not in header:
         raise ValueError('Missing "alg" in header')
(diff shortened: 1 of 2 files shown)


janwijbrand:
This PR is based on https://github.com/lepture/authlib/pull/390. 

Please let me know what I can do to make the PR qualify for merge. Thanks!

bjmc:
You may already be doing this, but if you are allowing users to specify this algorithm, make sure you are very careful about discouraging (maybe even explicitly disallowing) the use of the `none` algorithm, which has been [a source of many vulnerabilities](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/) in JWT-related libraries.

I notice some places where `None` is set as a default value and special checks like `if not None` are involved. We need to make it hard for users or future developers of this library to accidentally shoot themselves in the foot with insecure tokens. :slightly_smiling_face:

janwijbrand:
> You may already be doing this, but if you are allowing users to specify this algorithm, make sure you are very careful about discouraging (maybe even explicitly disallowing) the use of the `none` algorithm, which has been [a source of many vulnerabilities](https://auth0.com/blog/critical-vulnerabilities-in-json-web-token-libraries/) in JWT-related libraries.
> 
> I notice some places where `None` is set as a default value and special checks like `if not None` are involved. We need to make it hard for users or future developers of this library to accidentally shoot themselves in the foot with insecure tokens. 🙂

Thanks for the heads-up. I wonder: the PR is for code that creates the token, but, the vulnerabilities you refer to seems to be about the verification of tokens, right? I do think indeed we should help developers to the right thing.

bjmc:
You're correct about create/verify, but this code that creates the token should make it difficult (or IMHO impossible) to create an unsecured token.

janwijbrand:
> You're correct about create/verify, but this code that creates the token should make it difficult (or IMHO impossible) to create an unsecured token.

I agree and thanks for confirming my understanding of your remark. I hope the reviewer can assess if my approach at least makes it not too easy to forget passing in an signing alg.

janwijbrand:
And thanks @lepture for accepting the PR!!

PR #417, opened 2022-01-04
Use anyio.Lock to support trio/anyio backends in httpx client integrations

### LINKED ISSUE #394: httpx integration: add support for other async backends
**Is your feature request related to a problem? Please describe.**

At the moment authlib's httpx integration only supports the default asyncio backend (i.e. usage of asyncio.Event in AsyncOAuth2Client). But httpx supports multiple backends (asyncio, trio, curio, anyio)[^1]  

As a user I'd expect a library claiming to integrate with another lib to support the same environments.

Straightforward integration with the other async backends (trio, curio, anyio). 
Preferably in the way httpx provides it.

The following code snippet could be how a user chooses to use the trio backend with the AsyncOAuth2Client.
At least it is the way how it is done in httpx.

```python
from authlib.integrations.httpx_client import AsyncOAuth2Client 
import trio
```

[^1]: https://www.python-httpx.org/async/#supported-async-environments

richardsheridan:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

As title. Fixes #394. The anyio package is already a transitive dependency of httpx and starlette, so the only thing preventing usage of a trio backend with the async client integrations is the `asyncio.Event` being used to guard token refreshes. `anyio.Event` does not have a `reset` method, so I took the liberty of refactoring the critical section to use a `Lock`. The result is more readable, but I did have to update one test.

I'm not comfortable enough with your test suite to add a trio backend test, so I haven't tried to make any docs changes in this PR that announce trio compatibility as a new feature in the next release. Such testing may require anyio's pytest plugins.

---

### DIFF
--- a/authlib/integrations/httpx_client/oauth2_client.py
+++ b/authlib/integrations/httpx_client/oauth2_client.py
@@ -1,6 +1,7 @@
-import asyncio
 import typing
+
 from httpx import AsyncClient, Auth, Client, Request, Response, USE_CLIENT_DEFAULT
+from anyio import Lock  # Import after httpx so import errors refer to httpx
 from authlib.common.urls import url_decode
 from authlib.oauth2.client import OAuth2Client as _OAuth2Client
 from authlib.oauth2.auth import ClientAuth, TokenAuth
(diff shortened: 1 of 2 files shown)


lepture:
Thanks

AdamGold:
I think there's a bug with the locking here? https://github.com/lepture/authlib/issues/673

richardsheridan:
The lock has is initialized at the same time as the earlier event, so the lock itself is not inducing any bug here. Whether or not the token (for which the lock protects the refresh behavior) spans multiple sessions is beyond my understanding of the code.

PR #421, opened 2022-01-19
Fragment

btimby:

- [ ] Bugfix
- [X] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

---

This allows a parameter `in_fragment` to be passed into `prepare_grant_uri()`, when enabled, this will place OAuth parameters into the URL fragment rather than the query string. Further, it modifies `add_params_to_uri()` so that the original fragment is left unmolested (not made part of the parameters).

This is needed since I am using Vuejs for my frontend, and making API calls to perform OAuth2 authorization. I am using an authorization url such as:

http://localhost:8000/#/authorize

Without the `in_fragment` param, you would receive the url:

http://localhost:8000/?response_type=...#/authorize

And without preserving the fragment, you would get:

http://localhost:8000/#/authorize?%2Fauthorize&response_type=...

With these changes, the URL is correctly:

http://localhost:8000/#/authorize?response_type=...

### DIFF
--- a/authlib/common/urls.py
+++ b/authlib/common/urls.py
@@ -94,7 +94,7 @@ def add_params_to_uri(uri, params, fragment=False):
     """Add a list of two-tuples to the uri query components."""
     sch, net, path, par, query, fra = urlparse.urlparse(uri)
     if fragment:
-        fra = add_params_to_qs(fra, params)
+        fra += '?' + add_params_to_qs('', params)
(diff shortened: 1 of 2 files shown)


lepture:
You are doing OAuth wrong. From the specification, you are not allowed to implement authorization server in this way.

PR #446, opened 2022-04-04
Allow passing claims_options in StarletteOAuth2App

GaLaXy102:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

---

This pull requests allows to put custom `claims_options` into the `StarletteOAuth2App`. This is needed for example in a multihomed Keycloak setup. Here, the users accesses Keycloak from a public URL (e.g. `https://keycloak.localhost` with a Reverse Proxy in front) which might not be reachable from the application (e.g. in a Docker Container). In that case, the program using `authlib` would access Keycloak directly using a private URL (e.g. `http://keycloak:8080`).
Here, the JWT `iss` would mismatch when retrieving the `user_info`, but the developer knows that it has two possible values: The public and the private URL. To suppress the validation error, the base function `AsyncOpenIDMixin.parse_id_token` allows setting custom `claims_options`, but in `StarletteOAuth2App`, that was impossible to do.

### DIFF
--- a/authlib/integrations/django_client/apps.py
+++ b/authlib/integrations/django_client/apps.py
@@ -75,12 +75,13 @@ class DjangoOAuth2App(DjangoAppMixin, OAuth2Mixin, OpenIDMixin, BaseApp):
                 'state': request.POST.get('state'),
             }
 
+        claims_options = kwargs.pop('claims_options', None)
         state_data = self.framework.get_state_data(request.session, params.get('state'))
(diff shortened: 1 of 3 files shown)


lepture:
Could you also add it for Django and Flask? I would like to keep them the same API.

GaLaXy102:
Done

lepture:
Thanks

PR #450, opened 2022-04-06
fix #444: supply state to session and client

### LINKED ISSUE #444: Confusing behavior with OAuth2Session and state not being checked
**Describe the bug**

In the [documentation](https://docs.authlib.org/en/latest/client/oauth2.html#fetch-token) for how to use OAuth2Session client, it says that by supplying state when instantiating the object, then state will be checked when making the `fetch_token` request. In addition, the [docstring](https://github.com/lepture/authlib/blob/v1.0.0/authlib/oauth2/rfc6749/parameters.py#L131) for `parse_authorization_code_response` says that state is a required parameter when state is present in the client authorization request, but the [code](https://github.com/lepture/authlib/blob/v1.0.0/authlib/oauth2/rfc6749/parameters.py#L154) doesn't enforce that. Instead, it skips the check for state unless the user explicitly passes the state kwarg into the call to `fetch_token`. This leads to misleading behavior, where state is not actually checked.

**Error Stacks**

None

**To Reproduce**
We know there is a Flask OAuth client, and our example below doesn't use it, but uses Flask to create an easy, reproducible example. In our real app, we are using OAuth2Session client and not using Flask.
```python
import flask
import authlib.integrations.requests_client

app = flask.Flask(__name__)

@app.route('/')
def index():
    client = _client()
    uri, _ = client.create_authorization_url(
        'https://github.com/login/oauth/authorize',
        '<your server ip address>:8000/auth-github-authorize'
    )
    return flask.redirect(uri)

@app.route('/auth-github-authorized')
def auth_github_authorized():
    # FIXME: Supplying state here doesn't make a difference. It isn't checked.
    client = _client(state='a totally made up state')
    client.fetch_token(authorization_response=flask.request.url)
    raise AssertionError('Should not have gotten here. State is invalid.')


def _client(state=None):
    return authlib.integrations.requests_client.OAuth2Session(
        '<your-github-oauth-key>',
        '<your-github-oauth-secret>',
        scope='user:email',
        state=state,
        token_endpoint='https://github.com/login/oauth/access_token',
    )

if __name__ == '__main__':
    app.run(host='0.0.0.0', port=8000)
```

**Expected behavior**

authlib.oauth2.rfc6749.errors.MismatchingStateException should be raised.

**Environment:**

 - OS: Fedora 32
 - Python Version: 3.7.2
 - Authlib Version: 1.0.0

rorour:
> on README.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/integrations/requests_client/oauth2_session.py
+++ b/authlib/integrations/requests_client/oauth2_session.py
@@ -56,6 +56,7 @@ class OAuth2Session(OAuth2Client, Session):
     :param revocation_endpoint_auth_method: client authentication method for
         revocation endpoint.
     :param scope: Scope that you needed to access user resources.
+    :param state: Shared secret to prevent CSRF attack.
     :param redirect_uri: Redirect URI you registered as callback.
     :param token: A dict of token attributes such as ``access_token``,
         ``token_type`` and ``expires_at``.
@@ -74,7 +75,7 @@ class OAuth2Session(OAuth2Client, Session):
     def __init__(self, client_id=None, client_secret=None,
                  token_endpoint_auth_method=None,
                  revocation_endpoint_auth_method=None,
-                 scope=None, redirect_uri=None,
+                 scope=None, state=None, redirect_uri=None,
                  token=None, token_placement='header',
                  update_token=None, **kwargs):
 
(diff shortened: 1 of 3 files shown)


lepture:
Thanks

PR #454, opened 2022-05-04
Document required configuration for automatic token refreshes

acdha:
The `token_endpoint` is a hard requirement for automatic token refreshes but I had to trace the call from `OAuth2Session` to `OAuth2Client` to find the expected value.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Other, please describe: documentation

---

### DIFF
--- a/docs/client/oauth2.rst
+++ b/docs/client/oauth2.rst
@@ -270,6 +270,19 @@ protected resources. In this case, we can refresh the token manually, or even
 better, Authlib will refresh the token automatically and update the token
 for us.
 
+Automatically refreshing tokens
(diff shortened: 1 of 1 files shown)


acdha:
The example immediately following this is also quite confusing since it's copied from the Django client and uses the models defined in that class. I wasn't sure whether it'd be better to provide a generic example or to have that link to the Django example instead.

acdha:
> Thanks. It is not a Django model, it is a pseudo model class.

Ah, okay. I was just thinking from the perspective of it not being an example which you could use without substantial extra work.

PR #465, opened 2022-06-03
fix #464 with align httpx_client AsyncOAuth2Client.stream to httpx._c…

### LINKED ISSUE #464: httpx_client.oauth2_client.AsyncOAuth2Client.stream signature does not match httpx._client.AsyncClient.stream
**Describe the bug**

The method `httpx._client.AsyncClient.stream` is an asynccontextmanager and can be used like this:
```
client = httpx.AsyncClient()
async with client.stream('GET', 'https://www.example.com/') as response:
    async for chunk in response.aiter_bytes():
        # do something
```

The method `httpx_client.oauth2_client.AsyncOAuth2Client.stream` is not a asynccontextmanager and must be used like this:
```
client = AsyncOAuth2Client(...)
async with await client.stream("GET", 'https://i.b') as stream:
    async for chunk in response.aiter_bytes():
        # do something
```
You can see there is an `await` required as `client.stream()` is not an asynccontextmanager but returns an Awaitable, which has to be awaited to get the asynccontextmanager from internally the called `httpx_client.oauth2_client.AsyncOAuth2Client.stream`.

This breaks interoperability of the clients, which I think should be the case. This also did change with 1.x.x.

**To Reproduce**

Will add a PR to this issue, where I fix it and where you can reproduce it with the tests.

**Expected behavior**

I have an application where I just want to use different clients, e.g. the normal AsyncClient because I use BasicAuth and AsyncOAuth2Client for OAuth2 but wit the sam logic, this is currently not possible.

**Environment:**

 - OS: Debian buster and OS X
 - Python Version: 3.10.x
 - Authlib Version: Authlib==1.0.1

bjoernmeier:
…lient.AsyncClient.stream

The method httpx._client.AsyncClient.stream is an asynccontextmanager
and httpx._client.AsyncClient.stream should as well so both clients
allow interchangeability.

> on README.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

---


lepture:
Thanks

PR #476, opened 2022-07-29
Added client_id parameter to AssertionClient

vilmar-hillow:
Per https://datatracker.ietf.org/doc/html/rfc7521#section-4.1,
client_id parameter, although optional, can still be passed
when using assertions as authorization grants. Adding a way to pass
that id to refresh token body.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---


lepture:
1. I didn't see `client_id` is optional in the doc.
2. You are always passing `client_id=None`

vilmar-hillow:
> 1. I didn't see `client_id` is optional in the doc.
> 2. You are always passing `client_id=None`

1. From linked section: "Authentication of the client is optional, as described in
   [Section 3.2.1](https://datatracker.ietf.org/doc/html/rfc7521#section-3.2.1) of OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)], and consequently, the
   "client_id" is only needed when a form of client authentication that
   relies on the parameter is used."
   
  One of the providers I'm working with uses the authorization grant routine with client id.

2. Good catch, fixed

azmeuk:
I would guess client_id is indeed optional:

[RFC7521 section 4.1](https://datatracker.ietf.org/doc/html/rfc7521#section-4.1):

> Authentication of the client is optional, as described in
   [Section 3.2.1](https://datatracker.ietf.org/doc/html/rfc7521#section-3.2.1) of OAuth 2.0 [[RFC6749](https://datatracker.ietf.org/doc/html/rfc6749)], and consequently, the
   "client_id" is only needed when a form of client authentication that
   relies on the parameter is used.

[RFC6749 section 3.2.1](https://datatracker.ietf.org/doc/html/rfc6749#section-3.2.1)

>    A client MAY use the "client_id" request parameter to identify itself
   when sending requests to the token endpoint.

PR #503, opened 2022-10-26
docs: CodeChallenge(required=True) only applies to public clients

ktosiek:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Other, please describe: documentation change

---

I've had to look through code to make sure, but it seems the CodeChallenge is only required when the client uses "none" authentication - which I understand as being a public client. This makes it explicit in the documentation.


lepture:
Thanks

PR #505, opened 2022-10-30
Fixed RFC7592 Dynamic Client Registration Management Protocol

### LINKED ISSUE #499: RFC7592 Dynamic Client Registration Management Protocol
The [RFC7592 Dynamic Client Registration Management Protocol](https://www.rfc-editor.org/rfc/rfc7592) seems to be [implemented in authlib](https://github.com/lepture/authlib/blob/169c7dcfc47478c8d55553cc95fb0f5578162b77/authlib/oauth2/rfc7592/endpoint.py) but there is no reference in the documentation for this spec, nor unit test, and the README indicate that the spec is not implemented.

Is some development needed, or is it just the documentation and the tests that are missing?

azmeuk:
This PR fixes/implements the RFC7592 spec. It brings those changes

- The Flask integration keep the request data when the request type is `PUT`
- rfc7592/endpoint.py implements the RFC7592 specification
- The endpoint is tested by tests/flask/test_oauth2/test_client_configuration_endpoint.py
- rfc7592/endpoint.py has a 100% coverage
- The implementation is documented by docs/specs/rfc7592.rst

I read RFC7592 and did my best to implement this, but I am not 100% confident to have understood or implemented things right. I think a second pair of eyes might help. I tried to make unit tests as explicit and short as I could, and insert quotes from the spec to illustrate the tests.

The `extract_client_metadata` method has been copied from RFC7591 (and I removed the `software_statement` part). Maybe this should be factorized? If so, how should this be? I was thinking of extracting the method in a `ClientRegistrationEndpointMixin` from which would inherit `ClientRegistrationEndpoint` and `ClientConfigurationEndpoint`.

There are breaking changes on some method signatures, but this spec was not documented anyways:

- `introspect_client` MUST be implemented
- `save_client` is renamed `update_client` and takes an additional `client_metadata` parameter
- `introspect_metadata` is not required anymore
- `revoke_access_token` takes a `token` parameter

I think the documentation could make clearer that a *registration access token* can be a different thing than a regular *access_token*, but I guess this is enough like this.

Fixes #499

### DIFF
--- a/authlib/oauth2/rfc7592/endpoint.py
+++ b/authlib/oauth2/rfc7592/endpoint.py
@@ -37,7 +37,7 @@ class ClientConfigurationEndpoint(object):
             # If the client does not exist on this server, the server MUST respond
             # with HTTP 401 Unauthorized and the registration access token used to
             # make this request SHOULD be immediately revoked.
-            self.revoke_access_token(request)
+            self.revoke_access_token(request, token)
             raise InvalidClientError(status_code=401)
 
         if not self.check_permission(client, request):
@@ -193,20 +193,20 @@ class ClientConfigurationEndpoint(object):
         Developers MUST implement this method in subclass::
 
             def authenticate_client(self, request):
-                return Client.query.get(request.data.get('client_id'))
+                client_id = request.data.get('client_id')
+                return Client.get(client_id=client_id)
 
         :return: client instance
         """
         raise NotImplementedError()
 
-    def revoke_access_token(self, request):
+    def revoke_access_token(self, token, request):
         """Revoke a token access in case an invalid client has been requested.
         Developers MUST implement this method in subclass::
 
-            def revoke_access_token(self, request):
-                request.credential.revoked = True
-                db.session.add(request.token)
-                db.session.commit()
+            def revoke_access_token(self, token, request):
+                token.revoked = True
+                token.save()
 
         """
         raise NotImplementedError()
@@ -227,8 +227,7 @@ class ClientConfigurationEndpoint(object):
         implement it in subclass, e.g.::
 
             def delete_client(self, client, request):
-                db.session.delete(client)
-                db.session.commit()
+                client.delete()
 
         :param client: the instance of OAuth client
         :param request: formatted request instance
@@ -241,8 +240,7 @@ class ClientConfigurationEndpoint(object):
 
             def update_client(self, client, client_metadata, request):
                 client.set_client_metadata({**client.client_metadata, **client_metadata})
-                db.session.add(client)
-                db.session.commit()
+                client.save()
                 return client
 
         :param client: the instance of OAuth client
--- a/authlib/oauth2/rfc7592/endpoint.py
+++ b/authlib/oauth2/rfc7592/endpoint.py
@@ -250,3 +250,9 @@ class ClientConfigurationEndpoint(object):
         """
 
         raise NotImplementedError()
+
+    def get_server_metadata(self):
+        """Return server metadata which includes supported grant types,
+        response types and etc.
+        """
+        raise NotImplementedError()
--- a/authlib/oauth2/rfc7592/endpoint.py
+++ b/authlib/oauth2/rfc7592/endpoint.py
@@ -12,7 +12,7 @@ from ..rfc7591 import UnapprovedSoftwareStatementError
 
 
 class ClientConfigurationEndpoint(object):
-    ENDPOINT_NAME = "client_configuration"
+    ENDPOINT_NAME = 'client_configuration'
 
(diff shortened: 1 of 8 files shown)


authlib/oauth2/rfc7592/endpoint.py
```diff
@@ -145,28 +189,66 @@ def authenticate_token(self, request):
         raise NotImplementedError()
 
     def authenticate_client(self, request):
+        """Read a client from the request payload.
+        Developers MUST implement this method in subclass::
+
+            def authenticate_client(self, request):
+                return Client.query.get(request.data.get('client_id'))
```
lepture, line 196: please use a framework-free style example:

```py
            def authenticate_client(self, request):
                return query_client(request.data.get('client_id'))
```

authlib/oauth2/rfc7592/endpoint.py
```diff
@@ -145,28 +189,66 @@ def authenticate_token(self, request):
...
+        :return: boolean
+        """
         raise NotImplementedError()
 
     def delete_client(self, client, request):
         """Delete authorization code from database or cache. Developers MUST
         implement it in subclass, e.g.::
 
             def delete_client(self, client, request):
-                client.delete()
```
lepture, line 164: please use framework-free style example, `client.delete()` is better.

lepture:
1. I tried to remove sqlachemy from authlib.
2. are you using some code formatting tool? most of the code are using `'` instead of `"`, please don't change them.

lepture:
Thanks for your contribution, I'll check the specification implementation later.

azmeuk:
I have added commits for simple quotes and framework independent examples. I added commits to the PR so you can review the differences, but if you prefer (and for future fixes) I can squash the commit and force push.

What do you think about using a pre-commit configuration to check and enforce the code style for authlib? We could take inspiration from the [wtforms pre-commit configuration](https://github.com/wtforms/wtforms/blob/master/.pre-commit-config.yaml). I can open a PR if you wish.

lepture:
@azmeuk I don't like the black code style. Let's keep it the current way.

For testing, I'm using `self.assertEqual` and other `self.assertXXX` from unittest module. These method would print better error stack than the raw `assert`.

docs/changelog.rst
```diff
@@ -15,6 +15,7 @@ Version 1.2.0
 - Use ``flask.g`` instead of ``_app_ctx_stack``, via :gh:`issue#482`.
 - Add ``headers`` parameter back to ``ClientSecretJWT``, via :gh:`issue#457`.
 - Always passing ``realm`` parameter in OAuth 1 clients, via :gh:`issue#339`.
+- Implemented RFC7592 Dynamic Client Registration Management Protocol, via :gh:`issue#499`.
```
lepture, line 18:
```
via :gh:`PR#505`.
```

tests/flask/test_oauth2/test_client_configuration_endpoint.py
```diff
@@ -0,0 +1,505 @@
...
+        db.session.add(client)
+        db.session.commit()
+        return client
+
+    def authenticate_client(self, request):
+        client_id = request.uri.split('/')[-1]
+        return Client.query.filter_by(client_id=client_id).first()
+
+    def revoke_access_token(self, request, token):
+        token.revoked = True
```
lepture, line 34: Do we need to add:

```
db.session.add(token)
db.session.commit()
```

authlib/oauth2/rfc7592/endpoint.py
```diff
@@ -70,66 +71,109 @@ def create_delete_client_response(self, client, request):
...
 
-        client = self.save_client(client, request)
+        client_metadata = self.extract_client_metadata(request)
+        client = self.update_client(client, client_metadata, request)
         return self.create_read_client_response(client, request)
 
+    def extract_client_metadata(self, request):
+        json_data = request.data.copy()
+        options = self.get_claims_options()
+        claims = self.claims_class(json_data, {}, options, self.get_server_metadata())
```
lepture, line 108: You need to define:

```
def get_server_metadata(self):
    raise NotImplementedError()
```

tests/flask/test_oauth2/test_client_configuration_endpoint.py
azmeuk: Tests were passing without this, but I suppose this is needed anyways.

lepture:
We are almost done. But there should be a get_server_metadata method implementation in the rfc documentation.

Thanks for your hard work.

azmeuk:
This is done, does this implementation example enough?

lepture:
Thanks.

PR #523, opened 2023-01-15
jws: correctly handle empty payload with JSON serialization

zeldovich:
Previously the code checked if the payload value was True when converted to bool by the if statement, but that conflates None (i.e., no payload field in the JSON serialization) and the empty string (a specific payload), because both are False when converted to bool.  The proper check (as verified by the added test case) is to check if the payload is None.

- [ X ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

---

- [ X ] You consent that the copyright of your pull request source code belongs to Authlib's author.


lepture:
@zeldovich Thanks for the fix.

PR #533, opened 2023-03-20
Starlette v0.26.0 support and authorize_redirect fix

vicchi:
Starlette `0.26.0` changed the return value of `url_for()` from `str` to `URL`; see https://github.com/encode/starlette/pull/1385 for context.

Passing a `redirect_uri` to `authorize_redirect` causes a `TypeError: cannot convert 'URL' object to bytes` exception to be raised if the URL has been generated via `url_for()`. This fix detects an instance of `URL` and casts it to a string, thus allowing authentication to proceed and generally making people happy.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/integrations/starlette_client/apps.py
+++ b/authlib/integrations/starlette_client/apps.py
@@ -1,3 +1,4 @@
+from starlette.datastructures import URL
 from starlette.responses import RedirectResponse
 from ..base_client import OAuthError
 from ..base_client import BaseApp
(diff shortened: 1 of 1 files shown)


vicchi:
@lepture That is indeed true, but my thinking was that just casting to a string by default _could_ introduce sutble gotchas _if_ Starlette's `url_for()` _wasn't_ being used to build the redirect URL. We can be sure that Starlette's `URL` class does have an `__str__` dunder (see https://github.com/encode/starlette/blob/62b5b6042a39289ed561580c251c233250c3c088/starlette/datastructures.py#L166 for context) but we can't be sure there's nothing else at play here, hence me being cautious and checking that this is an instance of `URL` before casting.

More than happy to be told I'm being overly cautious here; it wouldn't be the first time.

vicchi:
@lepture Thanks! 😄

PR #534, opened 2023-03-23
docs: Update openID client userinfo usage

LudvigHz:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Other, please describe: Documentation

The documentation on obtaining userinfo through the OpenID userinfo endpoint seems to be incorrect,
as the `userinfo` method both does not expect positional arguments, and the request should in all
cases require a token?

I would suggest changing the function signature to include the `token` kwarg explicitly, as it is in
the `request` method. Let me know and I'll update this PR to include that as well.

---


lepture:
Thanks

PR #554, opened 2023-06-23
HTTPX: add send method

lspgn:
Hello,
Thank you for this library.
In some cases documented by [HTTPX](https://www.python-httpx.org/advanced/#request-instances), when using directly the `send` method, the bearer token is not passed.
This change adds a method with a similar signature to `request` and `stream` with an auth parameter.

The only thing I have not fully tested is `self.ensure_active_token(self.token)` as it's in a non async function.

Thank you

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/integrations/httpx_client/oauth2_client.py
+++ b/authlib/integrations/httpx_client/oauth2_client.py
@@ -104,6 +104,17 @@ class AsyncOAuth2Client(_OAuth2Client, httpx.AsyncClient):
             method, url, auth=auth, **kwargs) as resp:
             yield resp
 
+    def send(self, req, withhold_token=False, auth=USE_CLIENT_DEFAULT, **kwargs):
+        if not withhold_token and auth is USE_CLIENT_DEFAULT:
+            if not self.token:
+                raise MissingTokenError()
+
+            self.ensure_active_token(self.token)
+
+            auth = self.token_auth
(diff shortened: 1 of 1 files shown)


authlib/integrations/httpx_client/oauth2_client.py
```diff
@@ -104,6 +104,17 @@ async def stream(self, method, url, withhold_token=False, auth=USE_CLIENT_DEFAUL
             method, url, auth=auth, **kwargs) as resp:
             yield resp
 
+    def send(self, req, withhold_token=False, auth=USE_CLIENT_DEFAULT, **kwargs):
+        if not withhold_token and auth is USE_CLIENT_DEFAULT:
+            if not self.token:
+                raise MissingTokenError()
+
+            self.ensure_active_token(self.token)
```
lspgn, line 112: Replacing with the following fixes the issues but not sure about the solution
```python
            loop = asyncio.get_running_loop()
            asyncio.ensure_future(self.ensure_active_token(self.token), loop=loop)
```

lepture:
@lspgn thanks for your work. Maybe you need to add a test case for `send` method.

PR #570, opened 2023-08-26
end support for python 3.7

azmeuk:
Python 3.7 support stopped two months ago
https://endoflife.date/python

I removed tests for python 3.7 and applied [pyupgrade](https://pypi.org/project/pyupgrade/) --py38-plus to update the syntax for the recent python versions.


lepture:
wow, thanks

Issue #427, opened 2022-02-13
RFC9068: JWT access token


azmeuk:
Currently we can use `OAUTH2_ACCESS_TOKEN_GENERATOR` to customize access token generation, so this is not too hard to generate JWTs by re-using mechanisms in [generate_id_token](https://github.com/lepture/authlib/blob/169c7dcfc47478c8d55553cc95fb0f5578162b77/authlib/oidc/core/grants/util.py#L60).

However the recent [RFC9068](https://www.rfc-editor.org/rfc/rfc9068.html) standardize the claims the access token should carry.

I suggest making helpers in authlib to produce those tokens:

- helper to generate JWT access token in IdP
- helper to validate JWT access tokens for resource servers
- helpers to validate JWT access token in the IdP introspection endpoint
- token revocation endpoint should reject requests for JWT tokens, and return an [unsupported_token_type](https://datatracker.ietf.org/doc/html/rfc7009#section-2.2.1) error

What do you think?

lepture:
@azmeuk How about adding it to `oauth2.rfc9068` package? Can you send a PR for this feature?

azmeuk:
I won't probably have time to spent on this feature in a near future.

azmeuk:
My ideas on how this can be implemented:

- Refactor the `BearerTokenGenerator` part. Instead of passing custom methods via `OAUTH2_ACCESS_TOKEN_GENERATOR` or `OAUTH2_REFRESH_TOKEN_GENERATOR`, let the developers build their own classes that inherits from `BearerTokenGenerator` and build their own `access_token_generator`, `refresh_token_generator` and `expires_generator` if they want. Implement `AuthorizationServer.register_token_generator` to make the authorization server use the custom class.
- Implement a `BearerJWTTokenGenerator` class that inherits from `BearerTokenGenerator` and that builds JWT access tokens. This should be used by IdP to generate JWT access tokens.
- Implement a `BearerJWTTokenValidator` class that inherits from `BearerTokenValidator`, that decodes and validates JWT access tokens. This should be used by resource servers to validate JWT access tokens.

I wonder if we need a `JWTTokenIntrospectionEndpoint`, inheriting from `IntrospectionEndpoint` and returning token information about JWT access tokens for say resource servers that don't implement RFC9068 yet.

What do you think @lepture?

lepture:
Sound good to me. And you can use joserfc for JWT purposes. I will replace the authlib.jose module with joserfc.

PR #576, opened 2023-09-05
register several endpoints of a kind

azmeuk:
I am currently tackling #427 and I realize it would be easier for me if one could register several endpoints of a kind. For instance one `JWTAccessTokenIntrospectionEndpoint` that would read the introspection data from the JWT access token, and a regular `IntrospectionEndpoint` for the other kind of tokens.

With this patch, if `JWTAccessTokenIntrospectionEndpoint` would meet a plain-text access token, it would raise a `ContinueIteration` exception and pass to the `IntrospectionEndpoint`. Same if the token_hint is a refresh token for instance (RFC9068 is only about access tokens).

This would also help for the revokation endpoint.

What do you think?

### DIFF
--- a/authlib/common/errors.py
+++ b/authlib/common/errors.py
@@ -57,3 +57,7 @@ class AuthlibHTTPError(AuthlibBaseError):
         body = dict(self.get_body())
         headers = self.get_headers()
         return self.status_code, body, headers
+
+
+class ContinueIteration(AuthlibBaseError):
+    pass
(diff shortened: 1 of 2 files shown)


azmeuk:
An alternative to this would be to implement `register_introspection_endpoint` and `register_revokation_endpoint` with an optional `token_type` parameter, in the fashion of `register_token_generator`.
@lepture

azmeuk:
We can merge this now so the branch diff will be easier to read.

lepture:
Merged.

Issue #583, opened 2023-09-20
Delete/remove client method (flask client)


lcdunne:
I can only speak for the Flask client as that is what I'm using. It would be great if the `oauth.register(...)` method had an `oauth.remove(client_name)`.

**Is your feature request related to a problem? Please describe.**
I wanted to write a test for something and it involved registering an oauth2 (flask) client for testing purposes. After the tests complete, I want a clean way to remove that testing client from the oauth object, but I cannot see a way to do this. 

Without it, running the test poses the following issues. (1) If I create the client in my test with a fixed name (e.g. 'Test Client'), it will return the first client from the oauth registry for all subsequent tests, instead of creating a new client each time. But if I create the client with a unique name for each test (e.g. 'Test Client 1', 'Test Cleint 2', ...), the `oauth` registry will grow for as many clients are registered until the server restarts. Obviously the latter is the expected behaviour, but a `.remove_client` method seems missing

**Describe the solution you'd like**
Just like with `oauth.register(...)`, maybe something like `oauth.remove_client(client_name)` would be nice.

**Describe alternatives you've considered**
As far as I can tell, we might be able to do this by simply editing the internal `oauth._clients` and `oauth._registry` dictionaries. For example `del oauth._clients[my_test_client_name]; del oauth._registry[my_test_client_name]` but I don't know if this fully accomplishes the goal. It definitely doesn't feel nice.

lepture:
I think it is a good feature, feel free to send a pull request. The method should be added in every clients.

lcdunne:
Great, I'll take a look.

> The method should be added in every clients

Could you please explain a bit more what you mean by this? The `BaseOAuth` class looks like it is inherited by the clients, so I assumed it would be enough to just add the method there.

PR #585, opened 2023-09-23
Add remove_client method to base_client.registry.BaseOAuth

lcdunne:

- [ x ] Feature

Feature to support removing a client from the `oauth` instance (see [this issue](https://github.com/lepture/authlib/issues/583)).

---

- [ x ] You consent that the copyright of your pull request source code belongs to Authlib's author.

### DIFF
--- a/authlib/integrations/base_client/registry.py
+++ b/authlib/integrations/base_client/registry.py
@@ -86,6 +86,16 @@ class BaseOAuth:
         self._registry[name] = (overwrite, kwargs)
         return self.create_client(name)
 
+    def remove_client(self, name):
+        if name not in self._registry:
+            raise KeyError(f"Client {name} not found in registry.")
+
+        if name not in self._clients:
+            raise KeyError(f"Client {name} not found in clients.")
+        
+        del self._registry[name]
+        del self._clients[name]
+
     def generate_client_kwargs(self, name, overwrite, **kwargs):
         fetch_token = kwargs.pop('fetch_token', None)
         update_token = kwargs.pop('update_token', None)


lepture:
Hi, thanks. It would be great if you can add a test case for this method.

lcdunne:
Sure I'll give that a go. Is there a development build that installs all dependencies needed for testing? And any other information on how you are running the tests?

azmeuk:
The whole testing process is automated with tox. Just run tox, and this will create environments for every python version supported, and a few dependency variants, and this will run pytest on those environments.

andersnauman:
Can I give a suggestion to change the method name to `unregister` instead so the naming convention is kept in the same family so to say. Would also suggest the code to be more forgiven for odd behaviors (unlikely with this code but nonetheless a great failsafe).

Code suggestion:
```
    def unregister(self, name):
        err = None
        if name not in self._registry:
            err = f"Client {name} not found in registry."
        else:
            del self._registry[name]            

        if name not in self._clients:
            err = f"Client {name} not found in clients."
        else:
            del self._clients[name]

        if err:
            raise KeyError(err)
```

The method `create_client` first checks if `self._clients` contains `name` and if the `unregister` method do not finish deleting everything, you could end up with a half-deleted state where `create_client` returns a client that the registry does not have.

azmeuk:
> Can I give a suggestion to change the method name to unregister instead so the naming convention is kept in the same family so to say.

The *register* naming is not ideal anyways as it could be confused with registration like defined in [OAuth 2.0 Dynamic Client Registration Protocol](https://www.rfc-editor.org/rfc/rfc7591).

In the long run I think keeping `remove_client` but move `register` to `add_client` would be a better idea.

authlib/integrations/base_client/registry.py
```diff
@@ -86,6 +86,16 @@ def register(self, name, overwrite=False, **kwargs):
...
 
+    def remove_client(self, name):
+        if name not in self._registry:
+            raise KeyError(f'Client {name} not found in registry.')
+
+        if name not in self._clients:
+            raise KeyError(f'Client {name} not found in clients.')
+
+        del self._registry[name]
+        del self._clients[name]
```
Jamim, line 97: _Hello @lcdunne,_

This might be implemented in a bit cleaner and safer way.
```suggestion
        missing = []

        if not self._registry.pop(name, None):
            missing.append['registry']
        if not self._clients.pop(name, None):
            missing.append['clients']

        if missing:
            raise KeyError(f'Client {name} not found in f{" nor ".join(missing)}.')
```

_Best regards!_

Jamim (changes requested):

codespearhead:
Can you [link this PR to that issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) ( #583 ), [like this](https://github.com/springdoc/springdoc-openapi/pull/2540)?

codespearhead:
Also, can you [link this PR to that issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue) ( #583 ), [like this](https://github.com/springdoc/springdoc-openapi/pull/2540)?

Issue #524, opened 2023-01-16
RFC 8707: Resource Indicators for OAuth 2.0


azmeuk:
The [RFC8707](https://www.rfc-editor.org/rfc/rfc8707) details how a client can indicate to the authorization server on which resource server(s) the token will be used. This allows the authorization server to adapt the token claims depending on which resource server(s) it is intended, like for instance the `aud` and `scope` claims.

I suggest implementing this spec in authlib.

@lepture what do you think? Would you have implementation suggestions to tackle this?

lepture:
This should be implemented together with #427

azmeuk:
No. Interested in working on a PR?

gtadamson:
I encountered this issue while implementing RFC 8693 token exchange.

Both [[RFC 8693 §2.1](https://www.rfc-editor.org/rfc/rfc8693.html#section-2.1)] and [[RFC 8707 §2](https://www.rfc-editor.org/rfc/rfc8707.html#section-2)] allow a token request to contain multiple `resource` parameters:

```text
resource=https://api-one.example
resource=https://api-two.example
```

However, `AsyncOAuth2Client.fetch_token()` does not appear to provide a supported way to send repeated parameters.

I would expect to be able to pass the resources as a list:

```python
await client.fetch_token(
    token_endpoint,
    grant_type="urn:ietf:params:oauth:grant-type:token-exchange",
    resource=[
        "https://api-one.example",
        "https://api-two.example",
    ],
)
```

and have Authlib encode each value as a separate `resource` form parameter. I also tried passing a pre-encoded body containing repeated resource parameters, but `fetch_token()` does not preserve them either.

Could `fetch_token()` support list or tuple values for repeatable token-request parameters?

PR #586, opened 2023-09-26
RFC9068 implementation

### LINKED ISSUE #427: RFC9068: JWT access token

azmeuk:
This PR implements RFC9068 and fixes #427.

It implements:
- a token generator that builds JWT access tokens
- an introspection endpoint that introspects JWT access tokens
- a resource protector that validate authorizations against a JWT access token
- a revocation endpoint that refuses to revoke JWT access tokens

Everything is documented, and coverage is almost 100%, except on one part I will directly comment on the code.

I have left some comments for thing to be implemented with #524

Some notes about the implementation:
- resource protectors can take other arguments than `scope`, since RFC9068 introduces the `groups`, `entitlements` and `roles` claims to validate accesses
- `register_endpoint` can take objects, and not just classes
- I have not used joserfc yet. I thought it would be better to migrate the whole authlib library at once.

### DIFF
--- a/BACKERS.md
+++ b/BACKERS.md
@@ -103,5 +103,11 @@ Jeff Heaton
 </a><br>
 Birk Jernström
 </td>
+<td align="center">
+<a href="https://github.com/yaal-coop">
+<img src="https://avatars.githubusercontent.com/u/7792703?v=4" alt="Yaal Coop" width="48" height="48">
+</a><br>
+Yaal Coop
+</td>
 </tr>
 </table>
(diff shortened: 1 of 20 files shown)


lepture:
@azmeuk Is it ready for reviewing? If not, you can convert it to draft at first.

azmeuk:
Yes, this is ready for review!

PR #592, opened 2023-11-03
Make sure 'code' returns None instead of crashing if key is missing.

andersnauman:
> on README.

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

---

---

Instead of having the library users keeping track of which requests.args that are needed, change the behavior to not crash if variables requirements are not met. 

Example code before change:
```
authorize = Blueprint('authorize', __name__)

@authorize.route('/<client_name>')
def oauth(client_name):
    """ Verify oauth """
    args = request.args
    if None in (args.get("code"), args.get("state")):
        abort(500)
    client = _oauth.create_client(client_name)
    if not client:
        abort(404)
    token = client.authorize_access_token()
```

Example code after change:
Now the correct raise will occur which we can check for as expected:
```
from authlib.integrations.base_client.errors import MismatchingStateError
authorize = Blueprint('authorize', __name__)

@authorize.route('/<client_name>')
def oauth(client_name):
    """ Verify oauth """
    client = _oauth.create_client(client_name)
    if not client:
        abort(404)

    try:
        token = client.authorize_access_token()
    except MismatchingStateError:
        abort(500)
```

### DIFF
--- a/authlib/integrations/flask_client/apps.py
+++ b/authlib/integrations/flask_client/apps.py
@@ -85,12 +85,12 @@ class FlaskOAuth2App(FlaskAppMixin, OAuth2Mixin, OpenIDMixin, BaseApp):
                 raise OAuthError(error=error, description=description)
 
             params = {
-                'code': request.args['code'],
+                'code': request.args.get('code'),
                 'state': request.args.get('state'),
             }
         else:
             params = {
-                'code': request.form['code'],
+                'code': request.form.get('code'),
                 'state': request.form.get('state'),
             }
 


lepture:
It will not crash, Flask will raise a 400 bad request for this case, unless Flask has changed this behavior.

andersnauman:
You are correct that Flask will catch the crash and respond accordingly.
If we look into the context of the library, it still crashes. And while this might not pose a fatal error for the running process, it still makes the code behave unpredictable and stop further actions.

This can easily become a religious discussion, but I would suggest that any library should never crash but to return (or at least raise an error) a None state or any other way of giving a signal that something went wrong. The library user should not be forced to understand the underlying code just to run it in a safe matter.

Please also note that this code-change have already been made in other parts of your library.  E.g [Here](https://github.com/lepture/authlib/blob/ad13ae11b5c7f0eb1620c81917f285293c02081a/authlib/integrations/django_client/apps.py#L69C49-L69C49)

codespearhead (approved):

PR #594, opened 2023-11-16
Fix encode_client_secret_basic to match RFC 6749

### LINKED ISSUE #595: Client ID containing a colon doesn't work with OAuth2's client_secret_basic authentication method
Client ID containing a colon doesn't work with OAuth2's client_secret_basic authentication method.

The colon in the Client ID is currently sent as is, confusing the basic authentication parser in the server.

Prilkop:
Added url encoding of client_id and client_secret in encode_client_secret_basic per RFC 6749:
https://datatracker.ietf.org/doc/html/rfc6749#section-2.3.1

This fixes the unsupported situation where there is a colon character in the client_id or in the client_secret.

> on README.

- [x] Bugfix - fixes #595
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

---

### DIFF
--- a/authlib/oauth2/auth.py
+++ b/authlib/oauth2/auth.py
@@ -1,4 +1,5 @@
 import base64
+from urllib.parse import quote
 from authlib.common.urls import add_params_to_qs, add_params_to_uri
 from authlib.common.encoding import to_bytes, to_native
 from .rfc6749 import OAuth2Token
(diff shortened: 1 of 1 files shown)


lepture:
@Prilkop reverted, it caused too many bugs.

You can `quote` your client_id and client_secret at first.

adamrimon:
I agree with that I can quote the client_id and client_secret in my own code, but that's clearly against the RFC (requiring both to be url-encoded).
The bugs I saw were of people using previous version along with the new one - of course that could lead to a bug as this change fixes the previous incompatible wrong behavior (that is incompatible with other, compliant, servers).

PR #637, opened 2024-03-28
fix: OIDC "login" prompt should be "login" even if user authenticated

frankie567:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

---

---

OpenID Connect specification states the following about the `prompt=login` parameter:

> The Authorization Server SHOULD prompt the End-User for reauthentication. If it cannot reauthenticate the End-User, it MUST return an error, typically login_required.

Ref: https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest

In the current implementation, if `end_user` is present, the `login` prompt is ignored and set to `None`. We should instead keep this prompt so the end-developer can force a re-authentication of the user.


codespearhead (approved):
Good catch!

azmeuk:
Thank you for your contribution!

Issue #567, opened 2023-07-24
The expires_in function needs to have a timedelta to avoid tokenExpiry errors for milliseconds


pghole:
**Describe the bug**

I am using the OAuth2session object 

```
client = OAuth2Session(client_id=client_id, client_secret=client_secret, token_endpoint=token_url, grant_type='client_credentials')
client.fetch_token(token_url)
client.get(<MY_PROTECTED_URL>)
```
 Here, the library behavior is that the token gets automatically refreshed if that has expired. Refer https://github.com/lepture/authlib/blob/master/authlib/oauth2/client.py#L257 

However, the function which checks the token expiry https://github.com/lepture/authlib/blob/master/authlib/oauth2/rfc6749/wrappers.py#L13 , simply checks the expiry time with the current time . Because of this we are missing some corner cases, where the token is about to expire in few milliseconds/seconds and when the API call to the protected url is made, it gives error in authentication.

`JWT expired at 2023-06-20T13:16:42Z. Current time: 2023-06-20T13:16:42Z, a difference of 105 milliseconds.  Allowed clock skew: 0 milliseconds."
`
**Error Stacks**

`JWT expired at 2023-06-20T13:16:42Z. Current time: 2023-06-20T13:16:42Z, a difference of 105 milliseconds.  Allowed clock skew: 0 milliseconds."
`

**To Reproduce**

A minimal example to reproduce the behavior:

While the exact replication is not possible here as the request is failing by few milliseconds.

```
client = OAuth2Session(client_id=<client_id>, client_secret=<client_secret>, token_endpoint=<token_url>, grant_type='client_credentials')
client.fetch_token(<token_ur>l)
client.get(<MY_PROTECTED_URL>)
```

**A clear and concise description of what you expected to happen.**

Even if the token got expired by few milliseconds, the library should be able to handle such cases by obtaining a new token.
Instead of https://github.com/lepture/authlib/blob/master/authlib/oauth2/rfc6749/wrappers.py#L17 , we should be adding a small timedelta . For eg - even if the token is going to expire in next 60 seconds, refresh that still.

**Environment:**

 - OS: Linux
 - Python Version: 3.6
 - Authlib Version: 1.1.0

**Additional context**

There should be some timedelta introduced in the function , so that we can avoid facing issues where API requests fail by few milliseconds. Here, we can add logic to show that token has expired , let's say 30-60 seconds prior to its actual expiry.

princekhunt:
[...]

PR #643, opened 2024-04-11
Expose leeway in clients

michalismeng:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Other, please describe:

Issue #567 introduced a "leeway" parameter for proactive token refreshing. Expose this parameter in clients (e.g., the `requests` client) to allow configuring it by the library's users.

---

### DIFF
--- a/authlib/integrations/httpx_client/oauth2_client.py
+++ b/authlib/integrations/httpx_client/oauth2_client.py
@@ -58,7 +58,7 @@ class AsyncOAuth2Client(_OAuth2Client, httpx.AsyncClient):
                  revocation_endpoint_auth_method=None,
                  scope=None, redirect_uri=None,
                  token=None, token_placement='header',
-                 update_token=None, **kwargs):
+                 update_token=None, leeway=60, **kwargs):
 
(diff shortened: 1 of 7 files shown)


authlib/integrations/httpx_client/oauth2_client.py
```diff
@@ -58,7 +58,7 @@ def __init__(self, client_id=None, client_secret=None,
                  revocation_endpoint_auth_method=None,
                  scope=None, redirect_uri=None,
                  token=None, token_placement='header',
-                 update_token=None, **kwargs):
+                 update_token=None, leeway=60, **kwargs):
```
lepture, line 61: why default to `60`?

michalismeng: Because that's the default we've used at the lower-level function as well: https://github.com/lepture/authlib/blob/421732d50cf1f956aff4afa6a2a53f1da2b725f7/authlib/oauth2/rfc6749/wrappers.py#L13

Other than that, I don't have a strong opinion.

PR #651, opened 2024-05-30
OAuth2Request should access `grant_type` through `data`

apvd:
The `grant_type` property of the `OAuth2Request` class should refer to the `data` collection (consistent with the other properties `response_type`, `client_id`, etc. and not through the base `form`. Without this change the following grant setting will not function:

```
TOKEN_ENDPOINT_HTTP_METHODS = ["GET"]
```

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [X] No

---

### DIFF
--- a/authlib/oauth2/rfc6749/requests.py
+++ b/authlib/oauth2/rfc6749/requests.py
@@ -76,7 +76,7 @@ class OAuth2Request:
 
     @property
     def grant_type(self) -> str:
-        return self.form.get('grant_type')
+        return self.data.get('grant_type')
 
     @property
     def redirect_uri(self):


codespearhead:
I don't think that should be the case, as explained here: https://github.com/lepture/authlib/issues/658 .

codespearhead:
I don't so, as explained here: https://github.com/lepture/authlib/issues/658 .

apvd:
RFC frowns upon GET requests for token endpoint.
https://datatracker.ietf.org/doc/html/rfc6749#section-3.2

PR #662, opened 2024-07-19
OAuth2Client: use correct auth method for token introspection

AdamWill:
When token introspection was introduced in 6f5d19a, using the code that previously only handled token revocation, the new `_handle_token_hint` method that does the work for both `introspect_token` and `revoke_token` kept using
`self.revocation_endpoint_auth_method` unconditionally if no `auth` was passed in with the introspect or revoke request. This seems to be wrong, introspecting a token should use the `token_endpoint_auth_method`.

This leaves the fallback to `revocation_endpoint_auth_method` in `_handle_token_hint` because adjusting its signature to make `auth` compulsory would be awkward, but it's not expected ever to be used.

- [X] Bugfix

- [X] No

---

### DIFF
--- a/authlib/oauth2/client.py
+++ b/authlib/oauth2/client.py
@@ -299,6 +299,8 @@ class OAuth2Client:
 
         .. _`RFC7009`: https://tools.ietf.org/html/rfc7009
         """
+        if auth is None:
+            auth = self.client_auth(self.revocation_endpoint_auth_method)
         return self._handle_token_hint(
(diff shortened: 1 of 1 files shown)


AdamWill:
As I work for Red Hat I cannot agree to "consent that the copyright of your pull request source code belongs to Authlib's author" without my employer's permission, but I think this change is too trivial to be copyrightable.

Aliblame (approved):

Issue #633, opened 2024-03-21
Need to be able to add headers when calling authorize_redirect()


securitypedant:
**Is your feature request related to a problem? Please describe.**

I am using your library with the Flask client for an OAuth 2.0 flow.

The resource I am using your library with is behind a security gateway that is expecting auth in the HTTP headers. This is fine when I call...

token = oauth.myapp.authorize_access_token(headers=headers)

I can pass in HTTP headers for the auth, so that when the token is requested, it is authenticated by the HTTP gateway infront of my token issuer.

However, when I call...

return oauth.myapp.authorize_redirect(redirect_uri)

It looks for the server metadata, and gets it from a URL I specify in the registration, i.e.

server_metadata_url='https://dash.staging.mycompany.com/.well-known/jwks.json'

and the authorize_redirect() makes the HTTP request, but fails. I need a way to pass HTTP headers into this function as well.

**Describe the solution you'd like**

I would like to be able to run...

return oauth.myapp.authorize_redirect(redirect_uri, headers=headers)

Where headers contains HTTP authentication data.

**Describe alternatives you've considered**

I tried passing in the headers, hoping that **kwargs would be passed through and at some point the headers respected. But it didn't work.

**Additional context**

None

tianjing-li:
+1 I am also interested in having this feature

lepture:
A PR is welcome.

PR #710, opened 2025-02-20
Allow passing custom headers to metadata request in authorize_redirect

AlexandrosMaragkakis:
This pull request implements a feature to allow custom HTTP headers to be passed to the metadata request made by `authorize_redirect()`. 

Previously (#633), when calling `authorize_redirect(redirect_uri, headers=headers)`, the custom headers were not forwarded to the GET request for server metadata, causing a 401 Unauthorized error when the endpoint was protected by a security gateway.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---


azmeuk:
Hello. Thank you for your contribution. Would you consider adding some unit tests?

AlexandrosMaragkakis:
Hello. Unfortunately I don't have any experience in unit testing. Having said that, I am willing to give it a try. Perhaps you could suggest a file from the tests of the project, to serve as an example for me?

azmeuk:
You could take inspiration from [this test](https://github.com/lepture/authlib/blob/da87c8b2ec35af9ddd3b621e2e8245102018f878/tests/clients/test_flask/test_oauth_client.py#L281), register a client with custom headers, and then check if the mocked `send` method is called with your custom headers.
If you are not familiar with [mock](https://docs.python.org/3/library/unittest.mock.html), I guess the subject is pretty well covered on the internet.

authlib/integrations/base_client/sync_app.py
```diff
@@ -338,6 +341,11 @@ def create_authorization_url(self, redirect_uri=None, **kwargs):
         :param kwargs: Extra parameters to include.
         :return: dict
         """
+        headers = kwargs.pop("headers", None)
+        if headers:
+            if self.client_kwargs is None:
+                self.client_kwargs = {}
```
azmeuk, line 347: Is this needed as `client_kwargs` is automatically initialized as an empty dict when a `None` value is passed?
https://github.com/lepture/authlib/blob/da87c8b2ec35af9ddd3b621e2e8245102018f878/authlib/integrations/base_client/sync_app.py#L217

AlexandrosMaragkakis: You are correct, thanks for pointing it out. I also cleaned the other changes a bit.

AlexandrosMaragkakis:
I appreciate the guidance. I'll get on with adding the tests.

PR #711, opened 2025-02-21
Update starlette_client load_config function throwing TypeError

James-DeLorenzo:
The old syntax throws `TypeError: dict.get() takes no keyword arguments`. I followed the syntax for the [flask load_config function](https://github.com/lepture/authlib/blob/main/authlib/integrations/flask_client/integration.py#L26) to fix the issue.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/integrations/starlette_client/integration.py
+++ b/authlib/integrations/starlette_client/integration.py
@@ -66,7 +66,7 @@ class StarletteIntegration(FrameworkIntegration):
         rv = {}
         for k in params:
             conf_key = f"{name}_{k}".upper()
-            v = oauth.config.get(conf_key, default=None)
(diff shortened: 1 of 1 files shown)


James-DeLorenzo:
> Aren't you expected to use a Config object instead of a dict?

my goodness, I am not. Going over the docs it doesn't seem to me like there is direct mention of using starlette Configs vs dicts. Maybe that could be improved? Even the [api reference for calling OAuth](https://docs.authlib.org/en/latest/client/api.html#module-authlib.integrations.starlette_client) doesn't mention using them.

Thanks for pointing out this error, I'll close this pr now.

PR #558, opened 2023-06-27
Restore behaviour in create_authorization_response call which previously accepted a OAuth2Request object as-is

TurnrDev:
**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No - rather, fixes one. Explaination [here](https://github.com/lepture/authlib/commit/c060dab824f14929d33a567880042b026ed11b1b#r119835558)

---

- [x] You consent that the copyright of your pull request source code belongs to Authlib's author.

PR #748, opened 2025-04-16
Deprecate passing non `OAuth2Request` objects in `create_authorization_request`

azmeuk:
This code was introduced in #558 as a backward compatibility fix from 0.12, but is untested and should probably be avoided in the long run.


azmeuk:
Ooops if mixed up my thoughts.

PR #749, opened 2025-04-18
Implement RFC9101 JWT secured authentication requests (JAR)

### LINKED ISSUE #723: RFC9101 JWT-Secured Authorization Request (JAR)
The [OIDC JWTRequest](https://openid.net/specs/openid-connect-core-1_0.html#JWTRequests) concept has been generalized in its own dedicated spec [RFC 9101 The OAuth 2.0 Authorization Framework: JWT-Secured Authorization Request (JAR)](https://www.rfc-editor.org/rfc/rfc9101.html).

Basically, it consists in making clients pass authentication request details in a JWT instead of a query string. The JWT is the passed as a `request` parameter, or hosted on a client endpoint which url is passed as a `request_uri` parameter.

It would be nice for Authlib to implement this spec.

### LINKED ISSUE #724: RFC9126 Pushed Authorization Requests (PAR)
It would be nice for Authlib to support [Pushed Authorization Requests](https://www.rfc-editor.org/rfc/rfc9126.html) described in RFC9126. This would be a complement to #723
Basically, it consists in letting clients push a JWT with the authentication request detail to the authentication server, and get an URL on which the authentication process can be performed.

azmeuk:
This pull request brings a few refactoring and the implementation for [RFC 9101
The OAuth 2.0 Authorization Framework: JWT-Secured Authorization Request (JAR)](https://www.rfc-editor.org/rfc/rfc9101.html).

To summarize, RFC9101 is about passing the authorization request payload in a JWT, called request object, instead of passing it in the request body or query string. The request object is passed in a `request` parameter or hosted at the client and the URI to the request object is passed in a `request_uri` parameter.

To achieve this I had to perform some refactoring, and deprecate some usage with planned removal for Authlib 1.7 (so we can release 1.6 including this PR).

- The hook mechanism has been reworked, and the `execute_hook` method is replaced by a `hooked` decorator. This helps catching a method result every time when several `return` calls are in the method. There is no deprecation for hooks as I considered it is Authlib private API.
- A `AuthorizationServer` extension mechanism is introduced, and the RFC9207 `IssueParameter` is now an authorization server extension instead of a grant extension. Using `IssueParameter` as a grant extension is tolerated until 1.7.
- The `create_authorization_response` takes a `grant` parameter. This is to avoid calling `get_consent_grant` twice, which can be costly when a network request is performed to reach the `request_uri` endpoint. The parameter is optional at the moment but will become mandatory with 1.7.
- The `OAuth2Request` has been split in another `OAuth2Payload` class. This help separate the request object and the request payload, so the rfc9101 extension can overwrite the request payload with the request object content. Properties like `request.data` are deprecated in favor of `request.payload.data` and will be removed in 1.7.

Fixes #723

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,7 @@ Generic, spec-compliant implementation to build clients and providers:
   - [RFC8414: OAuth 2.0 Authorization Server Metadata](https://docs.authlib.org/en/latest/specs/rfc8414.html)
   - [RFC8628: OAuth 2.0 Device Authorization Grant](https://docs.authlib.org/en/latest/specs/rfc8628.html)
   - [RFC9068: JSON Web Token (JWT) Profile for OAuth 2.0 Access Tokens](https://docs.authlib.org/en/latest/specs/rfc9068.html)
+  - [RFC9101: The OAuth 2.0 Authorization Framework: JWT-Secured Authorization Request (JAR)](https://docs.authlib.org/en/latest/specs/rfc9101.html)
   - [RFC9207: OAuth 2.0 Authorization Server Issuer Identification](https://docs.authlib.org/en/latest/specs/rfc9207.html)
 - [Javascript Object Signing and Encryption](https://docs.authlib.org/en/latest/jose/index.html)
   - [RFC7515: JSON Web Signature](https://docs.authlib.org/en/latest/jose/jws.html)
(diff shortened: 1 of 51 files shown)


lepture:
There are many conflicts.

azmeuk:
They are fixed!

lepture:
Since this PR will be included in `1.6`, deprecating those `request` properties in `1.7` is too early, how about change it to `1.8`?

azmeuk:
@lepture this is fixed

lepture (approved):

PR #754, opened 2025-04-25
OIDC userinfo endpoint support

### LINKED ISSUE #459: Provide a UserInfo endpoint implementation
I've succesfully implemented an OAuth2/OIDC provider using authlib. But now stumbled upon a issue where a clients is not able to actually use the id_token (in addition to the access token, it can only use one token at a time). This shouldn't be an issue, as there's always the userinfo endpoint that can be read using the access token (or so I tought). I looked into adding the userinfo endpoint, but cannot find a implementation in authlib.

I'd like there to be a (reference) implementation of the userinfo endpoint as described in [section 5.3 of the OpenID Connect Core 1.0 spec](https://openid.net/specs/openid-connect-core-1_0.html#UserInfo).

azmeuk:
This PR implements the `UserInfoEndpoint` endpoint.
It also adds a `filter` method that restricts `UserInfo` claims to match a given scope.
Fixes #459

### DIFF
--- a/authlib/oauth2/rfc6749/models.py
+++ b/authlib/oauth2/rfc6749/models.py
@@ -221,3 +221,19 @@ class TokenMixin:
         :return: boolean
         """
         raise NotImplementedError()
+
+    def get_user(self):
+        """A method to get the user object associated with this token::
+
+        def get_user(self):
+            return User.get(self.user_id)
+        """
+        raise NotImplementedError()
+
+    def get_client(self) -> ClientMixin:
+        """A method to get the client object associated with this token::
+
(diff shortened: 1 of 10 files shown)


lepture (approved):

azmeuk:
@lepture are you able to merge this PR? For some reasons the button in disabled to me:

Capture d’écran du 2025-05-18 21-45-39

lepture:
@azmeuk I can not merge it.

lepture (approved):

azmeuk:
Ok. I will create a new PR.

PR #768, opened 2025-05-01
Client decode

### LINKED ISSUE #764: Support for decoding existing access tokens
**Is your feature request related to a problem? Please describe.**

While there's support for decoding JWT `id_token`s as part of an authlib client using the `authorize_redirect` and `authorize_access_token` calls, there doesn't appear to be a method to automatically decode an `access_token` sent as part of an Authorization Bearer header using the configuration setup in the client. Unless I've missed something, this requires replicating existing code that exists as part of the `OpenIDMixin.parse_id_token` (or `AsyncOpenIDMixin`).

The creation of a function similar to `authorize_access_token`, that takes a token and uses the existing client configuration to decode it.

**Describe alternatives you've considered**

You can currently use some non-documented but public functions of the client to perform JWT decoding manually. For example:

```py
oauth = OAuth()
oauth.register(
    'example',
    client_id=CLIENT_ID,
    client_secret=CLIENT_SECRET,
    server_metadata_url=f'{PROVIDER_URL}/.well-known/openid-configuration'
)

example = oauth.example
# Fetch access token using client as example, but could be retrieved as part of a request header
token = await example.fetch_access_token(scopes=SCOPES, grant_type='client_credentials')

# Simplified code from AsyncOpenIDMixin.parse_id_token
metadata = await example.load_server_metadata()
claims_params = {
    'client_id': example.client_id,
    'access_token': token['access_token'],
}
claims_options = {
    'iss': {
        'values': metadata['issuer']
    }
}
alg_values = metadata.get('id_token_signing_alg_values_supported')

jwt = JsonWebToken(alg_values)
jwk_set = await example.fetch_jwk_set()
claims = jwt.decode(
    token['access_token'],
    key=JsonWebKey.import_key_set(jwk_set),
    claims_cls=CodeIDToken,
    claims_options=claims_options,
    claims_params=claims_params,
)
```

**Additional context**

I'd be happy to work on a proof-of-concept/pull request implementation for the Starlette/Async OAuth2 client as that's what I'm currently working with, which can hopefully be used as a template for other providers. I'd appreciate some guidance on where best to implement this if so, and definitely would need review to make sure it's implemented in a secure manner.

Aetylus:
> on README.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

---

Resolves #764

Notes:
* Currently only implements for the async Starlette client, but if implementation looks good, I can implement the sync version as well.
* Initial tests have some duplicated code, but I was hesitant to start adding utility functions just for this to avoid that.

### DIFF
--- a/authlib/integrations/starlette_client/apps.py
+++ b/authlib/integrations/starlette_client/apps.py
@@ -8,6 +8,8 @@ from ..base_client.async_app import AsyncOAuth2Mixin
 from ..base_client.async_openid import AsyncOpenIDMixin
 from ..httpx_client import AsyncOAuth1Client
 from ..httpx_client import AsyncOAuth2Client
+from ...jose import JsonWebToken, JsonWebKey
+from ...oauth2.rfc9068.claims import JWTAccessTokenClaims
 
 
 class StarletteAppMixin:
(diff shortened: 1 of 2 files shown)


authlib/integrations/starlette_client/apps.py
```diff
@@ -97,3 +99,35 @@ async def authorize_access_token(self, request, **kwargs):
...
+                # responsible for requesting the access token to begin with
+                "sub": {"essential": True},
+                "client_id": {"essential": True},
+                "iat": {"essential": True},
+                "jti": {"essential": True},
+            }
+
+        # RFC9068: authorization servers and resource servers must support RS256
+        if alg_values is None:
+            alg_values = ["RS256"]
```
azmeuk, line 122: The `alg` value should be in the JWT header, so I am not sure this is needed to guess here.

You should be able to achieve the same thing with:
```
from authlib.jose import jwt
jwt.decode(token, jwks)
```

Aetylus: I apologize if my understanding of JWT and Access Tokens is a little sparse, but is there no concern with accepting any algorithm in the header? I did some looking around and [this article](https://curity.io/resources/learn/jwt-best-practices/) suggests that you shouldn't necessarily trust the header, but unsure how accurate it is:

> When verifying or decrypting the token you should always check the value of this claim with a list of algorithms that your system accepts. This mitigates an attack vector where someone would tamper with the token and make you use a different, probably less secure algorithm to verify the signature or decrypt the token.

I'm fine with removing the line and assuming the header `alg` is safe to use however if that's what we want to do.

PR #770, opened 2025-05-13
fix(starlette_client): remove default= keyword from config.get calls

aliaksei-protchanka:
Authlib's Starlette integration was calling self.config.get(conf_key, default=None), which raises a TypeError when config is a plain dict. Changed all occurrences to use the positional form self.config.get(conf_key, None) so that both dicts and Config objects work seamlessly.

> on README.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No


azmeuk:
Thank you for your contribution.

I am not sure why, but your patch causes the CI to fail.
I can reproduce the error locally, without this patch the test suite stays green.

PR #786, opened 2025-07-23
Raise a MissingCodeError when code parameter is missing

lepture:
ref https://github.com/authlib/authlib/issues/777

> DO NOT SEND ANY SECURITY FIX HERE. Please read "Security Reporting" section
> on README.

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [x] Yes
- [ ] No

If yes, please describe the impact and migration path for existing applications:

This affect the OAuth 2 framework clients authorize access token flow.

---

- [x] You consent that the copyright of your pull request source code belongs to Authlib's author.

PR #794, opened 2025-08-08
fix(client): raise MissingCodeException when code parameter is missing

lepture:
According to https://github.com/authlib/authlib/issues/793, our previous fix doesn't work well. This is another fix.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Other, please describe:

- [ ] Yes
- [x] No

---

### DIFF
--- a/authlib/integrations/base_client/__init__.py
+++ b/authlib/integrations/base_client/__init__.py
@@ -1,6 +1,5 @@
 from .errors import InvalidTokenError
 from .errors import MismatchingStateError
-from .errors import MissingCodeError
 from .errors import MissingRequestTokenError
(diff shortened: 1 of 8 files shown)


lepture:
cc @kaorihinata @markhobson

azmeuk:
Ah, actually we already had a `MissingCodeException`. I did not see that either...
At least #786 brought us that unit test, and it validates the fix here (the test fails if I disable your fix):
https://github.com/authlib/authlib/pull/786/files#diff-32309580fe6223cf33fe92fee41783e290868760c386af863839e13abcd6c551R530-R556

Issue #315, opened 2021-02-09
RFC 9449 OAuth 2.0 Demonstrating Proof of Possession (DPoP)


agentydragon:
**Is your feature request related to a problem? Please describe.**

I've been trying to talk to Solid (http://solidproject.org) servers with Python. Unfortunately, currently Solid server only works with DPoP tokens.

Here's a few links about DPoP:

* Draft RFC: https://tools.ietf.org/html/draft-fett-oauth-dpop-04
* https://darutk.medium.com/illustrated-dpop-oauth-access-token-security-enhancement-801680d761ff
* https://curity.io/resources/architect/oauth/dpop-overview/

No Python auth library that I found yet supports DPoP, so I had to do it mostly manually (see https://gitlab.com/agentydragon/solid-flask).

**Describe the solution you'd like**

Probably extend the OAuth2 session class so that it can authenticate itself via DPoP. It's might not be very difficult. Required changes:

* When sending token request, generate a key pair (https://gitlab.com/agentydragon/solid-flask/-/blob/afa42465d4e1e7b9daeeef7d99109f230c39402b/solid_flask_main.py#L196)
* Add the DPoP header to the token request; the content is a JWT signed by the generated key, and it also includes info about request time, request URL and method (so, in this case, the auth endpoint and POST): https://gitlab.com/agentydragon/solid-flask/-/blob/afa42465d4e1e7b9daeeef7d99109f230c39402b/solid_flask_main.py#L59
* Then on further requests:
  * Also include the DPoP header for the request
  * And instead of `Authentication: Bearer [access token]`, send the access token prefixed with `Authentication: DPoP`

**Describe alternatives you've considered**

* Other libraries
* Writing this myself when talking to the Solid server

lepture:
@agentydragon this is a draft spec and current draft is already expired.

> This Internet-Draft will expire on 5 September 2020.

I would like to wait for a while to implement this feature.

wwwald:
Here is a link to the current draft spec: [https://datatracker.ietf.org/doc/html/draft-ietf-oauth-dpop-08](https://datatracker.ietf.org/doc/html/draft-ietf-oauth-dpop-08)

The Solid ecosystem is heavily, heavily based on this and a reference Python implementation would be great to boost Python implementations against Solid.

No chance of picking this up again?

lepture:
I'm currently working on Mistune v3. I'll take my open source time into this feature after the release of Mistune v3.

lepture:
[...]

PR #808, opened 2025-08-27
RFC-9449: DPoP

zachswasey:
This is a WIP to request feedback early on in the approach before I clean things up, write tests, and submit an official PR.

This implements [RFC-9449: OAuth 2.0 Demonstrating Proof of Possession (DPoP)](https://datatracker.ietf.org/doc/html/rfc9449), #315 .

I'm supporting ATProto (Bluesky) in my application, and their implementation requires DPoP and PAR (which I also have a change I'm testing that I'll put up as a WIP PR later). You can read more about their specific requirements [here](https://docs.bsky.app/docs/advanced-guides/oauth-client) for context.

Please ignore the requests/django integration at the moment, as I've only focused on HTTPX in FastAPI at the moment. I'll update the other integrations once the approach is more finalized.

To use in Client:
* A dev would instantiate a rfc9449.DPoPProof, and pass that to the OAuthClient.
* The session JWK is created internally, and passed back on the token to be persisted along with the token since it needs to remain the same for the lifespan of the auth session and used for all interactions after /authorize or /par.
* A JWK can also be passed initially if desired, taken off the token that's passed into the client (if available), or have the mechanism to generate the JWK overridden if something besides ES256 is needed.
* TokenAuth/ClientAuth support retrying the request when a new nonce is returned from the auth/resource servers.

I explored refactoring TokenAuth/ClientAuth to support multiple Auths (separating Authorization header/client_secret_* modifications, and the DPoP proof header modifications), but that turned out to be more complicated than I originally expected. This would've introduced a new CompositeAuth to handle the pairs: TokenAuth/DPoPAuth, ClientAuth/DPoPAuth. I'm not against further exploring this idea, if it's desired by the maintainers. For now, I've implemented DPoP Proof directly in TokenAuth/ClientAuth implementations.

To use in Server:
```
dpop_nonce_generator = DefaultDPoPNonceGenerator()
dpop_proof_validator = DPoPProofValidator(nonce_generator=dpop_nonce_generator)
dpop_extension = DPoP(proof_validator=dpop_proof_validator)
server.register_grant(MyAuthorizationCodeGrant, [dpop_extension])
server.register_grant(MyRefreshTokenGrant, [dpop_extension])

require_oauth = ResourceProtector()
require_oauth.register_token_validator(MyDPoPTokenValidator(proof_validator=dpop_proof_validator))
```

Feel free to give any comments or ask questions as necessary about the implementation. Once I've got go ahead for the approach I'll clean up, implement the further integrations, add tests, and put up a final PR.

Thanks!

TODO:
* [ ] Other integrations
* [ ] Tests
* [x] Code in proper rfc* locations
* [x] An authorization server MAY elect to issue access tokens that are not DPoP bound, which is signaled to the client with a value of Bearer in the token_type parameter of the access token response per [RFC6750]. For a public client that is also issued a refresh token, this has the effect of DPoP-binding the refresh token alone, which can improve the security posture even when protected resources are not updated to support DPoP.
* [x] Add `dpop_signing_alg_values_supported` to metadata, and check alg against it
* [x] dpop_jkt parameter
* [x] server support (validate proofs, generate nonces, appropriate errors)
* [ ] Add nonce timestamp checking to proof iat checking

### DIFF
--- a/authlib/oauth2/client.py
+++ b/authlib/oauth2/client.py
@@ -1,19 +1,11 @@
-import json
-
 from authlib.common.security import generate_token
 from authlib.common.urls import url_decode
-from authlib.jose import JsonWebKey
-
-from .auth import ClientAuth
-from .auth import TokenAuth
+from .auth import ClientAuth, TokenAuth
 from .base import OAuth2Error
-from .rfc6749.parameters import parse_authorization_code_response
-from .rfc6749.parameters import parse_implicit_response
-from .rfc6749.parameters import prepare_grant_uri
-from .rfc6749.parameters import prepare_token_request
+from .rfc6749.parameters import parse_authorization_code_response, parse_implicit_response, prepare_grant_uri, \
+    prepare_token_request
 from .rfc7009 import prepare_revoke_token_request
 from .rfc7636 import create_s256_code_challenge
-from .rfc9449.proof import DPoPProof
 
 DEFAULT_HEADERS = {
     "Accept": "application/json",
@@ -99,7 +91,7 @@ class OAuth2Client:
 
         self.dpop_proof = dpop_proof
         if self.dpop_proof:
-            self.dpop_proof.generate_jwk(token)
+            self.dpop_proof.generate_jwk(token, metadata.get("dpop_signing_alg_values_supported", None))
         self.token_auth = self.token_auth_class(token, token_placement, self, self.dpop_proof)
 
         self.update_token = update_token
@@ -150,8 +142,6 @@ class OAuth2Client:
     @token.setter
     def token(self, token):
         self.token_auth.set_token(token)
-        if self.dpop_proof:
-            self.dpop_proof.set_token(token)
 
     def create_authorization_url(self, url, state=None, code_verifier=None, **kwargs):
         """Generate an authorization URL and state.
@@ -436,8 +426,7 @@ class OAuth2Client:
                 error=token["error"], description=token.get("error_description")
             )
         if self.dpop_proof:
-            token["dpop_jwk"] = self.dpop_proof.jwk.as_dict(is_private=True)
-            token["dpop_nonces"] = self.dpop_proof.nonces
+            token["dpop_jwk"] = self.dpop_proof.get_jwk_as_dict()
         self.token = token
         return self.token
 
--- a/authlib/oauth2/client.py
+++ b/authlib/oauth2/client.py
@@ -180,6 +180,9 @@ class OAuth2Client:
             kwargs["code_challenge"] = create_s256_code_challenge(code_verifier)
             kwargs["code_challenge_method"] = self.code_challenge_method
 
+        if self.dpop_proof:
+            kwargs["dpop_jkt"] = self.dpop_proof.jwk.thumbprint()
+
         for k in self.EXTRA_AUTHORIZE_PARAMS:
             if k not in kwargs and k in self.metadata:
                 kwargs[k] = self.metadata[k]
--- a/authlib/oauth2/client.py
+++ b/authlib/oauth2/client.py
@@ -98,9 +98,9 @@ class OAuth2Client:
         self.code_challenge_method = code_challenge_method
 
         self.dpop_proof = dpop_proof
-        self.token_auth = self.token_auth_class(token, token_placement, self, self.dpop_proof)
         if self.dpop_proof:
-            self.dpop_proof.set_token(token)
+            self.dpop_proof.generate_jwk(token)
+        self.token_auth = self.token_auth_class(token, token_placement, self, self.dpop_proof)
 
         self.update_token = update_token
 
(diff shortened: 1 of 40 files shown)


zachswasey:
Whoops, I wasn't aware that a Draft PR wouldn't notify anybody to actually take a look. So, marked as Ready for Review, but again, still just WIP and looking for high level feedback before finalizing. Thanks!

azmeuk:
Hi. Thank you for your contribution. We had the notification even when the PR was in draft.
I'll try to review your work, but as I did not study the DPoP spec yet, it may take some days or weeks.

authlib/oauth2/rfc6749/grants/refresh_token.py
```diff
@@ -9,14 +9,10 @@
 
 import logging
 
-from ..errors import InvalidGrantError
```
zachswasey, line 12: I'll update my Pycharm local settings to keep separate lines instead of group. That'll fix these import changes.

authlib/oauth2/rfc6749/models.py
```diff
@@ -165,6 +165,16 @@ def get_scope(self):
         """
         raise NotImplementedError()
 
+    def get_dpop_jkt(self):
```
zachswasey, line 168: Mixin.

authlib/oauth2/rfc6749/models.py
```diff
@@ -241,3 +251,23 @@ def get_client(self):
...
+        """A method to get the token type associated with this token:
+
+        .. code-block::
+
+            def get_token_type(self):
+                return self.token_type
+        """
+        raise NotImplementedError()
+
+    def get_dpop_jkt(self):
```
zachswasey, line 265: Mixin.

authlib/oauth2/rfc6749/resource_protector.py
```diff
@@ -39,6 +38,31 @@ def scope_insufficient(token_scopes, required_scopes):
...
+                    realm=self.realm,
+                    extra_attributes=self.extra_attributes
+                )
+
+        saved_token_type = token.get_token_type()
+        if saved_token_type.lower() != self.TOKEN_TYPE:
+            raise InvalidTokenError(description=f"Access token is of type {saved_token_type}, but token type is {self.TOKEN_TYPE}",
+                                    token_type=saved_token_type,
+                                    realm=self.realm,
+                                    extra_attributes=self.extra_attributes)
```
zachswasey, line 64: Mixin, if possible. Though this was a bit more difficult to do, but I'll give it another look.

zachswasey:
Based on the discussions on #814 I'll likely refactor the server-side component a bit, so take the existing structure with a grain of salt.

zachswasey:
Refactored and addressed most of the comments I left above, and from ideas from the PAR #814 PR. The one remaining bit is in `rfc6749/resource_protector.py` and `rfc6749/requests.py`, but I can address that after overall architecture comments.

I haven't fully finished the httpx/requests integrations yet, but I refactored their Auth classes to remove code duplication and make it easier to add new extension Auths in the future via a CompositeAuth, and handle retrying for auth reasons.

Again, please give feedback on the overall approach before I address cleanup, tests, docs, etc. I added an overall server extension which automatically hooks into AuthorizationCodeGrant and the RefreshTokenGrant, so developers don't have to add the same extension multiple times (like in the original top-level comment).

```
dpop_nonce_generator = DefaultDPoPNonceGenerator()
dpop_proof_validator = DPoPProofValidator(nonce_generator=dpop_nonce_generator)
server.register_extension(DPoP(proof_validator=dpop_proof_validator)

require_oauth = ResourceProtector()
require_oauth.register_token_validator(MyDPoPTokenValidator(proof_validator=dpop_proof_validator))
```

authlib/oauth2/client.py
```diff
@@ -93,6 +96,11 @@ def __init__(
         self.code_challenge_method = code_challenge_method
 
         self.token_auth = self.token_auth_class(token, token_placement, self)
+        self.dpop_proof = dpop_proof
+        if self.dpop_proof:
+            self.dpop_proof.generate_jwk(token, metadata.get("dpop_signing_alg_values_supported", None))
+            self.dpop_auth = DPoPAuth(self.dpop_proof)
```
zachswasey, line 102: Add `None` above.

snarfed:
@zachswasey are you still interested in trying to get this merged? I'm very interested in using it, at least the server side. What are your thoughts here?

snarfed:
I have a working draft of this, tested against a number of Bluesky clients : https://github.com/authlib/authlib/compare/main...snarfed:authlib:rfc9449. I narrowed it down to server only for now, updated it to use joserfc, and made DPoP and `get_dpop_jkt` optional. I still need to do a pass to incorporate some outstanding review comments here, and otherwise clean it up, but it's a functional start.

snarfed:
@azmeuk @lepture what are your thoughts here? Do you all have a preferred path forward for DPoP?  Let me know if you think it's worth my time to polish up https://github.com/authlib/authlib/compare/main...snarfed:authlib:rfc9126 , incorporate feedback from this PR, etc, and re-submit it as a new PR, or if you prefer a different direction, or if this isn't a priority for authlib right now. I'm ok with whatever.

PR #813, opened 2025-09-01
Migrate tests to pytest paradigm

azmeuk:

This migrates all the tests from the unittest paradigm to the pytest paradigm:

- Use pytest fixtures instead of setUp/tearDown. This allowed a bit of factorization.
- Use test functions instead of TestCase classes.
- Long test files have been split.
- Django and Flask authorization server config loading part has been moved to dedicated `load_config` methods.

This makes the unit tests a lot more legible, and a bit shorter. There are still 748 tests as before.

Follow-up to #746

**Checklist**

---


tests/flask/test_oauth1/test_resource_protector.py
```diff
@@ -1,170 +1,178 @@
...
+    db.session.add(tok)
+    db.session.commit()
+
+
+@pytest.mark.parametrize("use_cache", [True, False])
+def test_invalid_request_parameters(app, test_client, use_cache):
+    create_resource_server(app, use_cache, use_cache)
+    url = "/user"
+
+    # case 1
```
bjmc, line 56: Suggestion: Instead of having `case 1... case 5` all in one test, use `pytest.mark.parametrize()` to separate these into individual test cases.

azmeuk: I like pytest parametrize, but in that particular case pytest it a bit overkill to me. There are 3 vars to parametrize (url, error, error_description) for a 4 lines test. Anyway, I think both solutions would be fine.
I did not put a lot of factorization effort into OAuth1, but feel free to open a PR if you want to work on this.

lepture:
You don't need to migrate tests for jose module. Since we are going to remove jose module from authlib, we will also remove its tests.

azmeuk:
> You don't need to migrate tests for jose module.

I know but that was a really quick thing to do.
Do you prefer that I restore the way it was?

PR #814, opened 2025-09-01
RFC-9126: PAR

zachswasey:
This is a WIP to request feedback early on in the approach before I clean things up, write tests, and submit an official PR.

This implements [RFC-9126: OAuth 2.0 Pushed Authorization Requests](https://datatracker.ietf.org/doc/html/rfc9126)

I'm supporting ATProto (Bluesky) in my application, and their implementation requires DPoP and PAR. You can read more about their specific requirements [here](https://docs.bsky.app/docs/advanced-guides/oauth-client) for context.

This is only the client-side code at the moment, but again just looking for overall feedback before cleaning up and adding tests.

It's not particularly complicated on the client-side to support, but working it into the existing framework and maintain backwards compatibility was a bit difficult which is mostly where feedback would be welcome. Since PAR accepts all the same values at /par as would've normally been sent to /authorize I re-utilized `authorize_params` for both use-cases, and added a separate `authorize_url_params` arg specifically for parameters that need to be sent to /authorize in both PAR and non-PAR use-cases. I couldn't think of a cleaner way to do this, or even a better name (always horrible at naming things!), so suggestions are welcome.

Server support has been added as well. For saving the generated `request_uri` I added it to the AuthorizationCode, but that complicates the existing `query_authorization_code()` and `save_authorization_code()` methods since they'd need to be expanded to include the state, request_uri, and expires_in. I'm not sure how y'all have handled breaking changes like that in the past, so feedback explicitly about how you'd like to handle that would be most useful.

TODO:
* [ ] Tests
* [x] Server-side support

### DIFF
--- a/authlib/integrations/base_client/async_app.py
+++ b/authlib/integrations/base_client/async_app.py
@@ -1,6 +1,8 @@
 import logging
 import time
 
+from authlib.common.security import generate_token
+
 from authlib.common.urls import urlparse
 
 from .errors import MissingRequestTokenError
@@ -90,28 +92,59 @@ class AsyncOAuth2Mixin(OAuth2Base):
         async with self._get_oauth_client(**metadata) as session:
             return await _http_request(self, session, method, url, token, kwargs)
 
-    async def create_authorization_url(self, redirect_uri=None, **kwargs):
+    async def create_authorization_url(self, redirect_uri=None, authorize_url_params=None, **authorize_params):
         """Generate the authorization url and state for HTTP redirect.
 
         :param redirect_uri: Callback or redirect URI for authorization.
-        :param kwargs: Extra parameters to include.
+        :param authorize_url_params: Extra parameters specifically for the authorize_url endpoint, regardless of method.
+        :param authorize_params: Extra parameters to include for authorization.
         :return: dict
         """
         metadata = await self.load_server_metadata()
+        if self.authorize_params:
+            authorize_params.update(self.authorize_params)
+
+        authorize_url_params = authorize_url_params or {}
+        if self.authorize_url_params:
+            authorize_url_params.update(self.authorize_url_params)
+
         authorization_endpoint = self.authorize_url or metadata.get(
             "authorization_endpoint"
         )
         if not authorization_endpoint:
             raise RuntimeError('Missing "authorize_url" value')
 
-        if self.authorize_params:
-            kwargs.update(self.authorize_params)
+        par_endpoint = self.pushed_authorization_url or metadata.get(
+            "pushed_authorization_request_endpoint"
+        )
+        par_required = metadata.get("require_pushed_authorization_requests")
+        if par_required and not par_endpoint:
+            raise RuntimeError('Missing "pushed_authorization_url" value')
 
         async with self._get_oauth_client(**metadata) as client:
-            client.redirect_uri = redirect_uri
-            return self._create_oauth2_authorization_url(
-                client, authorization_endpoint, **kwargs
-            )
+            if redirect_uri is not None:
+                client.redirect_uri = redirect_uri
+
+            if par_required:
+                rv, request_uri = await self._pushed_authorization_request(client, par_endpoint, **authorize_params)
+                url, _ = client.create_authorization_url(
+                    authorization_endpoint,
+                    request_uri=request_uri['request_uri'],
+                    **authorize_url_params)
+                rv['url'] = url
+                return rv
+            else:
+                authorize_params.update(authorize_url_params)
+                return self._create_oauth2_authorization_url(client, authorization_endpoint, **authorize_params)
+
+    async def _pushed_authorization_request(self, client, par_endpoint, **kwargs):
+        rv = {}
+        params = self._get_authorization_params(client, **kwargs)
+        rv.update(params)
+        kwargs.update(params)
+        request_uri, state = await client.pushed_authorization(par_endpoint, **kwargs)
+        rv["state"] = state
+        return rv, request_uri
 
     async def fetch_access_token(self, redirect_uri=None, **kwargs):
         """Fetch access token in the final step.
--- a/authlib/integrations/base_client/async_app.py
+++ b/authlib/integrations/base_client/async_app.py
@@ -120,6 +120,7 @@ class AsyncOAuth2Mixin(OAuth2Base):
         par_required = metadata.get("require_pushed_authorization_requests")
         if par_required and not par_endpoint:
             raise RuntimeError('Missing "pushed_authorization_url" value')
+        par_required = par_required or par_endpoint
 
         async with self._get_oauth_client(**metadata) as client:
             if redirect_uri is not None:
--- a/authlib/oauth2/rfc6749/authorization_server.py
+++ b/authlib/oauth2/rfc6749/authorization_server.py
@@ -152,7 +152,9 @@ class AuthorizationServer(Hookable):
         self._client_auth.register(method, func)
 
     def register_extension(self, extension):
-        self._extensions.append(extension(self))
+        if extension not in self._extensions:
+            extension(self)
+            self._extensions.append(extension)
 
     def get_error_uri(self, request, error):
         """Return a URI for the given error, framework may implement this method."""
@@ -294,6 +296,7 @@ class AuthorizationServer(Hookable):
         endpoints = self._endpoints[name]
         for endpoint in endpoints:
             request = endpoint.create_endpoint_request(request)
+            request.endpoint = endpoint
             try:
                 return self.handle_response(*endpoint(request))
             except ContinueIteration:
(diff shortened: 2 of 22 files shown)


azmeuk:
First of all, a big thank you for the effort you put in those specs. :bow: 
I gave a quick look.

On the server side, the main issue I can see is that there is some RFC9126 code in the RFC6749 module (for example the `create_pushed_authorization_response` method, or the `AuthorizationCodeMixin` methods (`get_code` etc.)). The Authlib design is to keep the RFC specific code in dedicated modules and make them as non-intrusive as can be.
To achieve this you can [register authorization server extensions](https://github.com/authlib/authlib/blob/bafecc468162babde7428d9829fc5b019c835a33/authlib/oauth2/rfc6749/authorization_server.py#L154) and[register grant extensions](https://github.com/authlib/authlib/blob/bafecc468162babde7428d9829fc5b019c835a33/authlib/oauth2/rfc6749/authorization_server.py#L210), and then use [hooks](https://github.com/authlib/authlib/blob/main/authlib/oauth2/rfc6749/hooks.py) to trigger custom code at the right moment. You can also provide mixins and except developers to inherit them.
If needed you could define new hooks in rfc6749. All those mechanisms might not be enough to fill your need, in that case please tell us and we will think on how to find a better solution.

> I'm not sure how y'all have handled breaking changes like that in the past, so feedback explicitly about how you'd like to handle that would be most useful.

Breaking changes could be managed if necessary with a proper deprecation period. However, Authlib should be usable with or without PAR. I am not yet sure that PAR induces breaking changes. I might be smooth if it used the things I mention in the comments.

authlib/oauth2/rfc8414/models.py
```diff
@@ -33,6 +33,8 @@ class AuthorizationServerMetadata(dict):
         "introspection_endpoint_auth_methods_supported",
         "introspection_endpoint_auth_signing_alg_values_supported",
         "code_challenge_methods_supported",
+        "pushed_authorization_request_endpoint",
+        "require_pushed_authorization_requests",
```
azmeuk, line 37: Please put the rfc9126 metadata in a dedicated class in `rfc9126/discovery.py`, like this is done [with rfc9101](https://github.com/authlib/authlib/blob/main/authlib/oauth2/rfc9101/discovery.py) for example. Developers will then be able to use it with heritage:

```python
class AuthorizationServerMetadata(
    rfc8414.AuthorizationServerMetadata,
    rfc9101.AuthorizationServerMetadata,
    rfc9126.AuthorizationServerMetadata,
):
    pass
```

I know that this is not well documented at the moment, and that rfc8414 features are limited.

azmeuk:
On the server-side, [RFC9126 §3](https://openid.net/specs/oauth-v2-jarm.html#name-client-metadata) client metada is missing. You can take inspiration of [rfc9101/registration.py](https://github.com/authlib/authlib/blob/main/authlib/oauth2/rfc9101/registration.py) for example.

authlib/oauth2/rfc6749/grants/authorization_code.py
```diff
@@ -290,15 +343,44 @@ def create_token_response(self):
...
+        """The method to generate "code" value for authorization code data.
         Developers may rewrite this method, or customize the code length with::
 
             class MyAuthorizationCodeGrant(AuthorizationCodeGrant):
                 AUTHORIZATION_CODE_LENGTH = 32  # default is 48
         """
         return generate_token(self.AUTHORIZATION_CODE_LENGTH)
 
-    def save_authorization_code(self, code, request):
+    def generate_request_uri(self) -> Tuple[str, int]:
```
azmeuk, line 354: Maybe these methods could go in a mixin class in the rfc9126 module?

authlib/oauth2/rfc6749/models.py
```diff
@@ -165,6 +165,30 @@ def get_scope(self):
         """
         raise NotImplementedError()
 
+    def get_code(self):
```
azmeuk, line 168: Same here, maybe these methods can be part of a dedicated mixin class in the rfc9126 module.

authlib/oauth2/rfc6749/authorization_server.py
```diff
@@ -329,6 +332,21 @@ def create_authorization_response(self, request=None, grant_user=None, grant=Non
         grant.execute_hook("after_authorization_response", response)
         return response
 
+    def create_pushed_authorization_response(self, request=None, grant_user=None, grant=None):
```
azmeuk, line 335: I guess for this you could use the [register_endpoint](https://github.com/authlib/authlib/blob/bafecc468162babde7428d9829fc5b019c835a33/authlib/oauth2/rfc6749/authorization_server.py#L217) mechanism. You can take inspiration of [oidc UserInfo endpoint](https://github.com/authlib/authlib/blob/bafecc468162babde7428d9829fc5b019c835a33/authlib/oidc/core/userinfo.py) for example.

azmeuk:
A detail: Mixing client code and server code in the same PR is totally fine. It would maybe be easier for the review if it was separated though (because one could be merged before the other one, and that would help focusing the discussions around dedicated business logic.)

authlib/oauth2/rfc8414/models.py
zachswasey: Gotcha! Yeah, I really only added it because, well, other members were there. But I also noticed that actually creating an endpoint for rfc8414 doesn't seem to be included anywhere, or utilizing the validations on the client side. I had thoughts about putting up another PR for that, and being able to actually reference the metadata from within to determine behavior (like PAR enabled, or the DPoP algorithms). I assume that was the intention at some point?

zachswasey:
> A detail: Mixing client code and server code in the same PR is totally fine. It would maybe be easier for the review if it was separated though (because one could be merged before the other one, and that would help focusing the discussions around dedicated business logic.)

Yeah, I didn't originally intend to do server code at all, but it's made testing a lot easier as I could test the DPoP and PAR features individually since Bluesky is the only public endpoint I knew of that supported them. The intention was always to just get some high level feedback on the best way to integrate into Authlib (which you've given so far!) because y'all take a more unique approach to structuring the library than other similar solutions, and then I'll clean up, add tests, and will definitely put up a separate client/server PR to be merged separately and with squashed commits, for more granular comments.

I'll tackle the suggestions later today!

authlib/oauth2/rfc8414/models.py
zachswasey: Perfect. I'll take a closer look at how that file handles general authorization server extensions and apply it to the other PAR code, same with the server metadata. Ideally the get_server_metadata() should be abstracted out from any extension and made more central so it doesn't have to be re-implemented every time, and 8414 section 3.2 explicitly allows other claims from other specs, but I'll follow the pattern in 9101 for this PR.

azmeuk: > Ideally the get_server_metadata() should be abstracted out from any extension and made more central so it doesn't have to be re-implemented every time

Yes, I had this in my though for some times now :) At some point this will probably be refactored/redesigned.

authlib/integrations/base_client/async_app.py
```diff
@@ -90,28 +92,60 @@ async def request(self, method, url, token=None, **kwargs):
...
 
-        if self.authorize_params:
-            kwargs.update(self.authorize_params)
+        par_endpoint = self.pushed_authorization_url or metadata.get(
+            "pushed_authorization_request_endpoint"
+        )
+        par_required = metadata.get("require_pushed_authorization_requests")
+        if par_required and not par_endpoint:
+            raise RuntimeError('Missing "pushed_authorization_url" value')
+        par_required = par_required or par_endpoint
```
azmeuk, line 123: A detail, I would have called this var `should_use_par` instead.

This makes me think that this enforces the usage of PAR if available, and this is probably a good security practice. However, are there situations when one might want to avoid using PAR even if it is available on the server?

zachswasey:
I refactored PAR to follow the same model as JAR (RFC9101), which made a whole lot more sense once I realized what that was doing. This, however, required a refactoring of JAR due to conflicts with both acting on `request_uri` and JAR raising exceptions early ending the flow. So I introduced a RequestURIExtension to allow multiple handlers (JAR and PAR right now) to plug in and reliably handle request_uris of different sources.

As a result of the refactoring of JAR I deprecated the old class (which was also erroneously named JWTAuthenticationRequest), and introduced the refactored JWTAuthorizationRequest. I also updated the discovery/registration a bit to be cleaner and better match the rest of the code-base. I believe ClientMetadataClaims also erroneously subclassed BaseClaims from JOSE, which doesn't make any sense, so fixed that too. It should be fixed in 7591 too, but I left that since it wasn't part of the code I was touching.

As before, please take a look at the high level approach and give feedback. Once we're aligned, I'll focus on cleaning up a few things (RequestURIExtension should catch exceptions, docs, etc), write tests, and push a new PR.

zachswasey:
Any chance I can get some feedback on this and #808 so I can close things out on my end?

azmeuk:
Sorry I got busy elsewhere. I'll have a look in the week.

azmeuk:
I did not find time last week, but I don't lose hope for this one :crossed_fingers:

azmeuk:
It took me some time to read rfc9216 and rfc9101 but now I better see the big picture and how the two specs work together.
The PR looks good, the refactoring you propose look pertinent, the implementation is clean, there is not much coupling between modules. Nice work :+1: 
The refactoring in rfc9101 adds some complexity, but I am not sure how else can the `request_uri` parameter be managed.

azmeuk:
@zachswasey are you still interested in moving this forward?

snarfed:
Hi all! I'm very interested in this PR. @zachswasey, are you still working on it? If not, @azmeuk would you consider finalizing and merging it without them? And, do I understand right that it includes both server and client side support?

snarfed:
I have an updated draft of this PR, rebased up to current main, with minor updates and fixes, eg joserfc support. I've tested it against a number of Bluesky clients. https://github.com/authlib/authlib/compare/main...snarfed:authlib:rfc9126

I still need to do a pre-review pass to clean it up and make sure it incorporates the outstanding feedback here, but it's working, which is a good start.

@azmeuk @lepture et al, any thoughts? Is it worth polishing that branch into a new PR? Do you all have other plans for PAR? Or is this not a priority for authlib right now? I'm open to anything.

PR #827, opened 2025-09-26
feat: support list params in prepare_grant_uri

lisongmin:

I want to access logto using authlib recently, and It requires multiple resource keys in the query string (&resource=...&resource=...) in oidc/auth endpoint. when i using (resource=...+...) format, it raise error.
This maybe a bug in logto. But it is great if we support setting a list parameter in authlib.

This PR supports convert a list parameter into query params

**Checklist**

- [x] If this PR is about a new feature, or a behavior change, you have updated the documentation accordingly.

---


authlib/oauth2/rfc6749/parameters.py
```diff
@@ -56,7 +56,12 @@ def prepare_grant_uri(
 
     for k in kwargs:
         if kwargs[k] is not None:
```
lepture, line 58: how about adding an alias:

```
value = kwargs[k]
```

PR #840, opened 2025-11-24
Fix: Handle unparseable expires_at by falling back to expires_in

### LINKED ISSUE #833: `expires_in` should be used instead of `expires_at` if `expires_at` cannot be parsed
**Describe the bug**

When an OAuth2/OIDC server returns `expires_at` as a string that cannot be parsed as an integer, the `OAuth2Token` initialization fails with a `ValueError`. This prevents token creation even when `expires_in` is available and could be used to calculate the expiration time. 

Since `expires_at` is not defined in the OAuth2 RFC and some OIDC servers may return it in non-standard formats, the library should gracefully handle unparseable values and fall back to `expires_in`.

**Error Stack**

```python
Traceback (most recent call last):
  File "authlib/oauth2/rfc6749/wrappers.py", line 9, in __init__
    params["expires_at"] = int(params["expires_at"])
ValueError: invalid literal for int() with base 10: '<non-integer-string-value>'
```

**Steps to reproduce**

A minimal example to reproduce the behavior:
```python
from authlib.oauth2.rfc6749.wrappers import OAuth2Token

# Simulate a token response where expires_at is a non-integer string
token_params = {
    "access_token": "sample_token",
    "expires_at": "2025-10-10T12:00:00Z",  # Some servers return datetime string
    "expires_in": 3600,
    "token_type": "Bearer"
}

# This will raise ValueError in the old implementation
token = OAuth2Token(token_params)
```

**Expected behavior**

The OAuth2Token should be created successfully by falling back to expires_in when expires_at cannot be parsed as an integer. The library should calculate expires_at as current_time + expires_in and continue normal operation without raising an exception.

**Environment**

- OS: Linux
- Python Version: 3.x
- Authlib Version: 1.6.5

ShaikAyesha17:
Fixes #833

When an OAuth2/OIDC server returns expires_at as a string that cannot be parsed as an integer, the OAuth2Token initialization now gracefully handles the ValueError/TypeError and falls back to using expires_in to calculate the expiration time.

This prevents token creation from failing when expires_at is in a non-standard format while expires_in is available.

**Checklist**

- [ ] The commits follow the [conventional commits](https://www.conventionalcommits.org) specification.
- [ ] You ran the linters with ``prek``.
- [ ] You wrote unit test to demonstrate the bug you are fixing, or to stress the feature you are bringing.
- [ ] You reached 100% of code coverage on the code you edited, without abusive use of `pragma: no cover`

---

### DIFF
--- a/authlib/oauth2/rfc6749/wrappers.py
+++ b/authlib/oauth2/rfc6749/wrappers.py
@@ -4,7 +4,14 @@ import time
 class OAuth2Token(dict):
     def __init__(self, params):
         if params.get("expires_at"):
-            params["expires_at"] = int(params["expires_at"])
+                        try:
+
+                                            params["expires_at"] = int(params["expires_at"])
+                                        except (ValueError, TypeError):
+                                                            # If expires_at cannot be parsed, fall back to expires_in
+                                                            if params.get("expires_in"):
+                                                                                    params["expires_at"] = int(time.time()) + int(params["expires_in"])
+
         elif params.get("expires_in"):
             params["expires_at"] = int(time.time()) + int(params["expires_in"])
         super().__init__(params)


bendavis78:
I have a PR w/ proper formatting, tests, etc here: https://github.com/authlib/authlib/pull/842

PR #843, opened 2025-12-11
Client `id_token_signed_response_alg` metadata takes precedence over server `get_jwt_config`

### LINKED ISSUE #806: `id_token_signed_response_alg` should take precedence to sign `id_token`
**Describe the bug**

This is a follow-up to #802 and #755.

Currently, to find the alg to use to sign id_tokens, Authlib:

- use alg is set in `get_jwt_config`_ if defined
- else use `client.id_token_signed_response_alg` if defined
- else use the default RS256

I think `client.id_token_signed_response_alg` should be the primary option, as this is an explicit per-client configuration, defined in the spec. `get_jwt_config()["alg"]` could be considered as a *custom default alg*.

This may be a breaking change, so I think this could be introduced in 1.8.

Maybe if the `get_jwt_config` thing is overhauled, this can go with a deprecation time in 1.7?

azmeuk:

When signing id_tokens, the client `id_token_signed_response_alg` metadata is used in priority.
Fixes #806.

Before this, the value defined in `get_jwt_config` was used in priority, and `id_token_signed_response_alg` was only used if `get_jwt_config` does not provide `alg`.

This should only be merged for 1.7

### DIFF
--- a/authlib/oidc/core/grants/code.py
+++ b/authlib/oidc/core/grants/code.py
@@ -83,9 +83,7 @@ class OpenIDToken:
 
         # Per OpenID Connect Registration 1.0 Section 2:
         # Use client's id_token_signed_response_alg if specified
-        if not config.get("alg") and (
-            client_alg := request.client.id_token_signed_response_alg
-        ):
+        if client_alg := request.client.id_token_signed_response_alg:
             config["alg"] = client_alg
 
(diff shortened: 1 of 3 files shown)


azmeuk:
Actually I will do it in a way that does not bring a breaking change.

PR #854, opened 2026-01-16
Fix `expires_at` behavior when its value is 0

### LINKED ISSUE #530: Wrong handling of 0 in expires_at
If the `expires_at` timestamp in a token is 0 the `is_expired()` function erroneously returns `None` instead of `True`.
The check should explicitly check for `None`.

https://github.com/lepture/authlib/blob/ee4337cf7c825349dd23870822a3cc7df123097f/authlib/oauth2/rfc6749/wrappers.py#L13-L17

azmeuk:
Fix the behavior when `expires_at=0`.
fix #530


lepture (approved):

lepture:
LGTM

PR #855, opened 2026-01-16
Make require_oauth parenthesis optional

### LINKED ISSUE #604: Allow the instance of ResourceProtector to be a decorator without an unnecessary call if we don't have any 'call' attribute, solution provided.
for example on this page of [documentation](https://docs.authlib.org/en/latest/flask/2/resource-server.html) we can see:

```python
@app.route('/user')
@require_oauth()
def user_profile():
    user = current_token.user
    return jsonify(user)

# or with None

@app.route('/user')
@require_oauth(None)
def user_profile():
    user = current_token.user
    return jsonify(user)
```
If we speak about transparency in coding, `@require_oauth()` is not an obvious practice.
More often, you can encounter `@require_oauth`.
Organizing the decorator for both cases — calling with and without attributes — is easy:

```python
    def __call__(self, *args, **kwargs):
        if args and callable(args[0]):
            return super().__call__()(*args, **kwargs)
        return super().__call__(*args, **kwargs)
```

azmeuk:
fix #604

### DIFF
--- a/authlib/integrations/django_oauth1/resource_protector.py
+++ b/authlib/integrations/django_oauth1/resource_protector.py
@@ -49,7 +49,7 @@ class ResourceProtector(_ResourceProtector):
         return req.credential
 
     def __call__(self, realm=None):
-        def wrapper(f):
+        def decorator(f):
             @functools.wraps(f)
             def decorated(request, *args, **kwargs):
                 try:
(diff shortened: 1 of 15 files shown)


lepture:
LGTM, but tests are missing

PR #861, opened 2026-02-14
fix: normalize resolve_client_public_key method

lepture:
If the method is used to resolve a client public key, it is renamed to `resolve_client_public_key`, and it will only take a `client` parameter.

### DIFF
--- a/authlib/oauth2/rfc7523/client.py
+++ b/authlib/oauth2/rfc7523/client.py
@@ -1,11 +1,16 @@
+from __future__ import annotations
+
 import logging
 
+from joserfc import jwk
 from joserfc import jws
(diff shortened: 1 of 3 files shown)


azmeuk:
I was wondering in term of design where the best place for resolving client keys would be, this could be in the modules as it is now, or in the client mixin. Since it will probably involve requests maybe the modules is the right place indeed.

lepture:
Do you mean we should have a client.get_public_key() method?

azmeuk:
Well, this is an option. What do you think?

lepture:
@azmeuk I think we can add the client method later. Maybe in next release.

PR #889, opened 2026-05-14
feat: Auto-propagate kid and alg from JWK to JWT header in RFC 7523 assertion signing

liudonggalaxy:

This is a feature implementation.

For the OAuth2 client credentials flow, when using `private_key_jwt` or `client_secret_jwt` authentication with a `JWK` that already contains `kid` and `alg`, the caller still has to extract those values from the key and pass them separately, even though the key itself is already passed to the function. This is cumbersome, especially in projects with many OAuth2 clients, where every call site has to repeat the same boilerplate. E.g.,

Before: 
```python
from joserfc.jwk import OKPKey, OctKey
 
 # Example 1: Ed25519 private key (private_key_jwt)
 ed25519_key = OKPKey.generate_key(crv="Ed25519")
 ed25519_key.update({"kid": "ed25519-key-1", "alg": "Ed25519"})
 
 # Manually extract kid and alg from the key we already pass
 token = private_key_jwt_sign(
     private_key=ed25519_key,
     client_id="my-client",
     token_endpoint="https://auth.example.com/token",
     alg=ed25519_key.get("alg"),               # redundant
     header={"kid": ed25519_key.get("kid")},   # redundant
 )
 
 # Example 2: HMAC shared secret (client_secret_jwt)
 hmac_key = OctKey.import_key(shared_secret)
 hmac_key.update({"kid": "hmac-key-1", "alg": "HS384"})
 
 # Same boilerplate for every call site in the project
 token = client_secret_jwt_sign(
     client_secret=hmac_key,
     client_id="another-client",
     token_endpoint="https://auth.example.com/token",
     alg=hmac_key.get("alg"),               # redundant
     header={"kid": hmac_key.get("kid")},   # redundant
 )
```

Solution: Automatically copy `kid` and `alg` from the `JWK` into the `JWT` header via a new helper `set_jwt_header_parameter_from_key(...)`. The key's value is an enforced constraint (joserfc raises if `header alg != key alg`), so it takes priority over any explicit value.

Priority order:

 1. `key.alg / key.kid` — highest priority (enforced by joserfc)
 2. Explicit `alg/kid` parameter — used when key has no `alg/kid`
 3. Function default — `HS256` for `client_secret_jwt_sign`, `RS256` for `private_key_jwt_sign`

After:
```python
from joserfc.jwk import OKPKey, OctKey
 
 # Example 1: Ed25519 private key
 ed25519_key = OKPKey.generate_key(crv="Ed25519")
 ed25519_key.update({"kid": "ed25519-key-1", "alg": "Ed25519"})
 
 # Just pass the key — kid and alg are read from it automatically
 token = private_key_jwt_sign(
     private_key=ed25519_key,
     client_id="my-client",
     token_endpoint="https://auth.example.com/token",
 )
 
 # Example 2: HMAC shared secret
 hmac_key = OctKey.import_key(shared_secret)
 hmac_key.update({"kid": "hmac-key-1", "alg": "HS384"})
 
 # Same simplification — no manual extraction needed
 token = client_secret_jwt_sign(
     client_secret=hmac_key,
     client_id="another-client",
     token_endpoint="https://auth.example.com/token",
 )
```

Fully backward compatible:
   - Keys without `alg/kid` → behavior unchanged, explicit params or defaults apply
   - Keys with `alg/kid` → they were already required to match (joserfc enforces this), so auto-propagation just removes the redundant boilerplate

**Checklist**

---

### DIFF
--- a/authlib/oauth2/rfc7523/assertion.py
+++ b/authlib/oauth2/rfc7523/assertion.py
@@ -1,9 +1,26 @@
 import time
+from typing import Any
 
 from joserfc import jwt
+from joserfc._rfc7517.models import BaseKey
 
 from authlib._joserfc_helpers import import_any_key
 from authlib.common.security import generate_token
+
+
+def set_jwt_header_parameter_from_key(
+    header: dict[str, Any], key: Any, parameter_name: str
+) -> None:
+    """
+    Copy a JWK parameter (e.g. alg, kid) from key into header.
+
+    The key's value is an enforced constraint and takes priority over any
+    value already present in header.
+    """
+    if isinstance(key, BaseKey):
+        parameter_value = key.get(parameter_name)
+        if parameter_value:
+            header[parameter_name] = parameter_value
 
 
 def sign_jwt_bearer_assertion(
@@ -17,13 +34,18 @@ def sign_jwt_bearer_assertion(
     header=None,
     **kwargs,
 ):
+    _key = import_any_key(key)
+
     if header is None:
         header = {}
     alg = kwargs.pop("alg", None)
     if alg:
         header["alg"] = alg
+    set_jwt_header_parameter_from_key(header=header, key=_key, parameter_name="alg")
     if "alg" not in header:
         raise ValueError("Missing 'alg' in header")
+
+    set_jwt_header_parameter_from_key(header=header, key=_key, parameter_name="kid")
 
     payload = {"iss": issuer, "aud": audience}
 
@@ -44,7 +66,7 @@ def sign_jwt_bearer_assertion(
     if claims:
         payload.update(claims)
 
-    return jwt.encode(header, payload, import_any_key(key), algorithms=[header["alg"]])
+    return jwt.encode(header, payload, _key, algorithms=[header["alg"]])
 
 
 def client_secret_jwt_sign(
--- a/authlib/oauth2/rfc7523/assertion.py
+++ b/authlib/oauth2/rfc7523/assertion.py
@@ -2,7 +2,10 @@ import time
 from typing import Any
 
 from joserfc import jwt
-from joserfc._rfc7517.models import BaseKey
+from joserfc.jwk import ECKey
+from joserfc.jwk import OctKey
+from joserfc.jwk import OKPKey
+from joserfc.jwk import RSAKey
 
 from authlib._joserfc_helpers import import_any_key
 from authlib.common.security import generate_token
@@ -17,7 +20,7 @@ def set_jwt_header_parameter_from_key(
     The key's value is an enforced constraint and takes priority over any
     value already present in header.
     """
-    if isinstance(key, BaseKey):
+    if isinstance(key, (OctKey, RSAKey, ECKey, OKPKey)):
         parameter_value = key.get(parameter_name)
         if parameter_value:
             header[parameter_name] = parameter_value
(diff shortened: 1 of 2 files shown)


lepture (approved):

authlib/oauth2/rfc7523/assertion.py
```diff
@@ -1,11 +1,28 @@
 import time
+from typing import Any
 
 from joserfc import jwt
+from joserfc._rfc7517.models import BaseKey
```
lepture: Better not import from a private module. You can use 

```python
from joserfc.jwk import OctKey, RSAKey, ECKey, OKPKey

if isinstance(key,(OctKey, RSAKey, ECKey, OKPKey)):
```

liudonggalaxy: Done

authlib/oauth2/rfc7523/assertion.py
```diff
@@ -1,11 +1,31 @@
...
+    """
+    Copy a JWK parameter (e.g. alg, kid) from key into header.
+
+    The key's value is an enforced constraint and takes priority over any
+    value already present in header.
+    """
+    if isinstance(key, (OctKey, RSAKey, ECKey, OKPKey)):
+        parameter_value = key.get(parameter_name)
+        if parameter_value:
+            header[parameter_name] = parameter_value
```
lepture, line 26: I think you should use `header.setdefault`, if the given header contains `alg` field, you should not use the `alg` value from the key.

liudonggalaxy: > if the given header contains alg field, you should not use the alg value from the key.

1. [client_secret_jwt_sign(...)](https://github.com/authlib/authlib/blob/bf39e2f239eeb39e4600770f93cab487e89596c6/authlib/oauth2/rfc7523/assertion.py#L51) uses `HS256` as the default algorithm. [private_key_jwt_sign(...)](https://github.com/authlib/authlib/blob/bf39e2f239eeb39e4600770f93cab487e89596c6/authlib/oauth2/rfc7523/assertion.py#L57) uses `RS256` as the default algorithm. These defaults override `key.alg`. As a result, the following usage is no longer supported. We have to manually provide the alg parameter to the function.
```
token = private_key_jwt_sign(
    private_key=ed25519_key,
    client_id="my-client",
    token_endpoint="https://auth.example.com/token",
)
```

Likewise, the following pattern is no longer valid:
```
with OAuth2Session(
    client_id,
    client_secret,
    token_endpoint_auth_method=ClientSecretJWT(token_endpoint),
    token_endpoint=token_endpoint,
    grant_type="client_credentials",
) as session:
```

Instead, the algorithm and headers must be specified explicitly:
```
with OAuth2Session(
    client_id,
    client_secret,
    token_endpoint_auth_method=ClientSecretJWT(
        token_endpoint,
        alg=client_secret["alg"],
        headers={"kid": client_secret.get("kid")},
    ),
    token_endpoint=token_endpoint,
    grant_type="client_credentials",
) as session:
```


2. There are three possible sources for the algorithm (alg):

- `header.alg`
- The function parameter `alg`
- `key.alg`

If either `header.alg` or the function parameter `alg` does not match `key.alg`, `joserfc` raises an exception and the function call fails.

In practice, the only way for users to avoid this failure is to ensure that `header.alg` matches `key.alg`. Therefore, `key.alg` effectively acts as a constraint on all other algorithm values and has the final authority over algorithm selection.

As a result, `key.alg` has the highest priority because it can reject `alg` values provided from any other source.

PR #892, opened 2026-05-15
feat: Include issued_at and expires_at in token response, respect server timestamps on client

liudonggalaxy:

This is a feature implementation

Add `issued_at` and `expires_at` to the bearer token response body, and teach the client-side `OAuth2Token` to prefer the server-provided `issued_at` over the local clock when computing expiration.

**Server side (`BearerTokenGenerator.generate()`)**
- Captures a single `time.time()` call and resolves `expires_in` **before** calling `access_token_generator()`, storing both as `self._issued_at` and `self._expires_in`
- `issued_at` and `expires_at` are included in the response whenever `expires_in` is present
- For JWT tokens, `JWTBearerTokenGenerator.access_token_generator()` reads `self._issued_at` and `self._expires_in` directly, guaranteeing `iat == issued_at` and `exp == expires_at` — no duplicate `time.time()` or `_get_expires_in()` calls

**Client side (`OAuth2Token`)**
- When the server provides `issued_at`, the client computes `expires_at` from it instead of from the local clock, avoiding clock-skew issues
- Invalid `issued_at` values (zero, negative, non-numeric) fall back to the client clock gracefully
- Falls back to `time.time()` when `issued_at` is absent (backward compatible)

**Persistence (`save_token` helpers)**
- The default `save_token` helpers in both the SQLAlchemy and Django integrations now filter the token dict to known model columns before persisting, so new response-only fields do not break existing schemas

**Checklist**

---

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,4 @@ venv/
 uv.lock
 .env
 coverage.xml
+authlib.iml
--- a/authlib/integrations/django_oauth2/authorization_server.py
+++ b/authlib/integrations/django_oauth2/authorization_server.py
@@ -54,7 +54,9 @@ class AuthorizationServer(_AuthorizationServer):
             user_id = request.user.pk
         else:
             user_id = client.user_id
-        item = self.token_model(client_id=client.client_id, user_id=user_id, **token)
+        fields = {f.name for f in self.token_model._meta.get_fields()}
+        token_data = {k: v for k, v in token.items() if k in fields}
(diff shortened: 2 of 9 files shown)


.gitignore
```diff
@@ -22,3 +22,4 @@ venv/
 uv.lock
 .env
 coverage.xml
+authlib.iml
```
liudonggalaxy, line 25: Ignore `IntelliJ IDE` project file

PR #893, opened 2026-05-15
feat: fallback to trying all JWKS keys when kid does not match

liudonggalaxy:

This is a feature implementation that adds a configurable fallback for OIDC ID token verification when the token's `kid` header does not match any key in the provider's JWKS.

**What is the current behavior?**

When `jwt.decode()` raises `InvalidKeyIdError`, the code refreshes the JWKS and retries. If the kid still does not match after refresh, `InvalidKeyIdError` is raised with no way to recover. This breaks integration with OIDC providers whose JWKS keys lack a `kid` field or use `kid` values that do not match the token header.

**What is the new behavior?**

On `InvalidKeyIdError`:
1. Refresh JWKS via `fetch_jwk_set(force=True)` to handle key rotation
2. Retry kid-based decode with refreshed keyset
3. If `should_try_all_keys()` returns `True`, try each key individually; otherwise re-raise `InvalidKeyIdError` (default, backwards-compatible)

Subclasses can override `should_try_all_keys()` to enable the brute-force fallback:
```python
class MyOAuth(OAuth):
    def should_try_all_keys(self):
        return True
```

**Checklist**

---

### DIFF
--- a/authlib/integrations/base_client/async_openid.py
+++ b/authlib/integrations/base_client/async_openid.py
@@ -8,6 +8,8 @@ from authlib.oidc.core import CodeIDToken
 from authlib.oidc.core import ImplicitIDToken
 from authlib.oidc.core import UserInfo
 
+from .sync_openid import try_all_keys
+
 __all__ = ["AsyncOpenIDMixin"]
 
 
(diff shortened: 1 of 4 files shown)


lepture:
Is there a real use case? I don't think it is a good idea go accept non-matched keys.

liudonggalaxy:
> Is there a real use case? I don't think it is a good idea go accept non-matched keys.

Yes.  Some identity providers do not include `kid` values in their JWKS responses. For example, `Palantir’s Multipass` JWKS endpoint may return keys like this:
```
{
    "keys": [
        {
            "kty": "EC",
            "crv": "P-256",
            "x": "t85AHwgZP8Sz_0Q0YZOr4kn4WxXrQeYa7OPDJwg2cnM",
            "y": "GHPmt7eNhuiffYymKabMppQwS-HipAbFyDuidZ5DoJ0"
        },
        {
            "kty": "EC",
            "crv": "P-256",
            "x": "cW__b1XeUUoXEzTMf7ty6rjy38HokQ7GGtMTL5ZC-1E",
            "y": "pLXB3SjRNZyTw8cQrF0mZQO7OjTASzrNOT0xsBKXZ-U"
        }
    ]
}
```
The `mozilla-django-oidc` library supports a similar [OIDC_VERIFY_KID](https://mozilla-django-oidc.readthedocs.io/en/stable/settings.html#OIDC_VERIFY_KID) settings to disable `kid` check for JWKS keys.

PR #895, opened 2026-05-18
feat: auto-fetch token for client_credentials grant in OAuth2Session/OAuth2Client

liudonggalaxy:

This is a feature implementation.

When using `OAuth2Session` (requests) or `OAuth2Client` (httpx) with the `client_credentials` grant type, the caller must explicitly call `fetch_token()` before making any resource request, even though all the required metadata (`token_endpoint`, `grant_type`, `client_id`, `client_secret`) is already configured on the session. This is inconsistent with `AssertionSession`, which transparently auto-fetches tokens via `ensure_active_token()`.

Before:

```python
from authlib.integrations.requests_client import OAuth2Session

sess = OAuth2Session(
    client_id="my-client",
    client_secret="my-secret",
    token_endpoint="https://auth.example.com/token",
    grant_type="client_credentials",
)

# Must manually fetch before any request — raises MissingTokenError otherwise
sess.fetch_token("https://auth.example.com/token")
resp = sess.get("https://api.example.com/resource")
```

After:

```python
from authlib.integrations.requests_client import OAuth2Session

sess = OAuth2Session(
    client_id="my-client",
    client_secret="my-secret",
    token_endpoint="https://auth.example.com/token",
    grant_type="client_credentials",
)

# Token is fetched automatically on first request
resp = sess.get("https://api.example.com/resource")
```

Solution: In `request()` (and `stream()` for httpx), when `self.token` is `None` and the configured `grant_type` is `"client_credentials"`, call `self.fetch_token()` automatically. Since `fetch_token()` already reads `token_endpoint` and `grant_type` from `self.metadata` when not explicitly provided, the call is simply `self.fetch_token()` with no arguments needed.

For non-`client_credentials` grant types, the existing `MissingTokenError` / `OAuthError` behavior is preserved unchanged.

Fully backward compatible — sessions that already call `fetch_token()` explicitly continue to work identically.

**Checklist**

- [x] You ran the linters with `prek`.

---

### DIFF
--- a/authlib/integrations/requests_client/oauth2_session.py
+++ b/authlib/integrations/requests_client/oauth2_session.py
@@ -6,7 +6,6 @@ from authlib.oauth2.auth import TokenAuth
 from authlib.oauth2.client import OAuth2Client
 
 from ..base_client import InvalidTokenError
-from ..base_client import MissingTokenError
 from ..base_client import OAuthError
 from ..base_client import UnsupportedTokenTypeError
 from .utils import update_session_configure
@@ -134,7 +133,5 @@ class OAuth2Session(OAuth2Client, Session):
         if self.default_timeout:
             kwargs.setdefault("timeout", self.default_timeout)
         if not withhold_token and auth is None:
-            if not self.token:
-                raise MissingTokenError()
             auth = self.token_auth
         return super().request(method, url, auth=auth, **kwargs)
--- a/authlib/integrations/httpx_client/oauth2_client.py
+++ b/authlib/integrations/httpx_client/oauth2_client.py
@@ -14,7 +14,6 @@ from authlib.oauth2.auth import TokenAuth
 from authlib.oauth2.client import OAuth2Client as _OAuth2Client
 
 from ..base_client import InvalidTokenError
-from ..base_client import MissingTokenError
 from ..base_client import OAuthError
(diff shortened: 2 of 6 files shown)


authlib/integrations/requests_client/oauth2_session.py
```diff
@@ -134,7 +133,5 @@ def request(self, method, url, withhold_token=False, auth=None, **kwargs):
         if self.default_timeout:
             kwargs.setdefault("timeout", self.default_timeout)
         if not withhold_token and auth is None:
-            if not self.token:
-                raise MissingTokenError()
```
liudonggalaxy, line 138: **Why the `if not self.token: raise MissingTokenError()` was removed from `requests_client/oauth2_session.py`'s `request()` method:**

The token check and auto-fetch is now handled automatically by the auth layer. When `request()` sets `auth = self.token_auth`, it creates an `OAuth2Auth` instance. During the request, `OAuth2Auth.__call__()` invokes `self.client.ensure_active_token(self.token)`, which:

1. If `token is None` and `grant_type="client_credentials"` → calls `self.fetch_token()` automatically
2. If `token is None` and no auto-fetch is possible → raises `MissingTokenError`
3. If token exists but is expired → refreshes it

So the explicit check in `request()` was redundant — `OAuth2Auth` already triggers `ensure_active_token()` which handles all cases.

Issue #694, opened 2024-12-19
Proxy can not be set for AsyncOAuth2Client


sebastian-heinz:
**Describe the bug**

AsyncOAuth2Client accept kwargs that are passed to the httpx.AsyncClient.

httpx.AsyncClient accept `proxy` as a kwarg.
however:
`client_kwargs = self._extract_session_request_params(kwargs)`

drops anything that is not in here:
HTTPX_CLIENT_KWARGS = [
    'headers', 'cookies', 'verify', 'cert', 'http1', 'http2',
    'proxies', 'timeout', 'follow_redirects', 'limits', 'max_redirects',
    'event_hooks', 'base_url', 'transport', 'app', 'trust_env',
]

`proxies` afaik is used in requests or other libraries to provide a dict of protocol/host, however httpx client seems to only expect `proxy` kwarg, which is a string with protocol://host format.

**Expected behavior**

passing `proxy="http://host.tld:8080"` as kwarg to `AsyncOAuth2Client` should be passed to the underlying httpx client

**Environment:**

 - OS: osx
 - Python Version: 3.12.7
 - Authlib Version: 1.3.2

lepture:
It seems httpx has changed its parameter.

lepture:
Released in [Version 1.4.1](https://github.com/lepture/authlib/releases/tag/v1.4.1)

PR #909, opened 2026-07-10
feat(client): use httpx2 instead of httpx

### LINKED ISSUE #904: Support httpx2 over httpx
pydantic has taken over httpx with a new package httpx2.  Have you considered migrating?

levinKaus:

A feature implementation of #904: switches the `authlib.integrations.httpx_client` integration from `httpx` to `httpx2` as its underlying HTTP client library. This PR has two commits with distinct purposes:

1. `feat(client): use httpx2 instead of httpx`, the actual dependency swap. This is a mechanical change with no new behavior: `httpx2`'s `Client`/`AsyncClient` API is a near 1:1 match for `httpx`'s (same constructor kwargs, `Auth`/`Request`/`Response`/`USE_CLIENT_DEFAULT`/`WSGITransport`). No new tests were added for this commit; the existing test suite for `httpx_client` and the Starlette integration was updated to import from `httpx2` and re-run against the real package (94 tests in `tests/clients/test_httpx` and `tests/clients/test_starlette`, all passing). The full `tests/clients` suite (206 tests, spanning the unrelated `requests`/Flask/Django integrations too) also passes, confirming no regression elsewhere.

2. `fix(client): remove dead app kwarg handling in sync httpx clients`, a small, unrelated cleanup found while verifying 100% coverage on the code touched by commit 1. The `app=`/`WSGITransport` shim in `OAuth1Client`, `OAuth2Client`, and `AssertionClient` turned out to be dead code, so I dug into the git history to understand why before touching it:
   - It was added in Dec 2024 ("Support httpx 0.28") to preserve the old `app=` convenience after httpx dropped it from `Client`/`AsyncClient` directly, for all six client classes (sync and async).
   - `"app"` was then removed from `HTTPX_CLIENT_KWARGS` in a Jan 2025 commit tied to #694, which silently broke the shim for sync and async alike (the kwarg was never extracted into the dict the shim later popped from).
   - A follow-up commit six days later removed the now-dead shim code, but only from the three **async** classes. The identical dead code in the three **sync** classes was left behind untouched. Given the async removal happened alongside an unrelated async-specific proxy-kwarg fix, this reads like an oversight rather than a deliberate decision to keep it for sync.
   - Since this dead branch was inherited unchanged into the `httpx2`-based code and was the only thing keeping `oauth1_client.py`, `oauth2_client.py`, and `assertion_client.py` under 100% coverage on the lines touched by commit 1, I removed it to match the cleanup already done on the async side, rather than add tests for a feature that never actually worked.

Yes, for commit 1. `authlib.integrations.httpx_client` now requires `httpx2` instead of `httpx`. Existing applications need to run `pip install httpx2` instead of `pip install httpx`. The public import path (`authlib.integrations.httpx_client`) and all class names (`OAuth1Client`, `OAuth2Client`, `AssertionClient`, `AsyncOAuth1Client`, `AsyncOAuth2Client`, `AsyncAssertionClient`, `OAuth1Auth`, `OAuth2Auth`, etc.) are unchanged, so no application code changes are required beyond the dependency swap in most cases.

However, applications that interact with `httpx` types directly around an Authlib client call will need to migrate those usages to `httpx2` as well:

- **Exception handling**: `httpx2` defines its own exception hierarchy (`httpx2.HTTPError`, `httpx2.TimeoutException`, `httpx2.ConnectError`, etc.), distinct from `httpx`'s. Code doing `except httpx.HTTPError: ...` around calls made through an Authlib client will silently stop catching those errors after upgrading.
- **Transports and mocking**: a custom `transport=` (e.g. `httpx.MockTransport`, a custom `httpx.BaseTransport` subclass, or tooling like `respx` that patches `httpx` internals) will not work with the new `httpx2`-based clients and needs an `httpx2` equivalent.

Authlib itself does not catch any `httpx`-specific exception types internally, so this only affects application code that references `httpx` types directly.

Commit 2 (the dead code removal) is **not** a breaking change. The code it removes was unreachable, so no application could have been relying on it. Behavior is identical before and after.

**Checklist**

- [x] If this PR is about a new feature, or a behavior change, you have updated the documentation accordingly.

---

### DIFF
--- a/authlib/integrations/httpx_client/assertion_client.py
+++ b/authlib/integrations/httpx_client/assertion_client.py
@@ -1,13 +1,13 @@
-import httpx2
-from httpx2 import USE_CLIENT_DEFAULT
-from httpx2 import Response
-
 from authlib.oauth2.rfc7521 import AssertionClient as _AssertionClient
 from authlib.oauth2.rfc7523 import JWTBearerGrant
 
 from ..base_client import OAuthError
+from ._compat import httpx2
 from .oauth2_client import OAuth2Auth
 from .utils import extract_client_kwargs
+
+USE_CLIENT_DEFAULT = httpx2.USE_CLIENT_DEFAULT
+Response = httpx2.Response
 
 __all__ = ["AsyncAssertionClient"]
 
--- a/authlib/integrations/httpx_client/assertion_client.py
+++ b/authlib/integrations/httpx_client/assertion_client.py
@@ -91,11 +91,6 @@ class AssertionClient(_AssertionClient, httpx2.Client):
         **kwargs,
     ):
         client_kwargs = extract_client_kwargs(kwargs)
-        # app keyword was dropped!
-        app_value = client_kwargs.pop("app", None)
-        if app_value is not None:
-            client_kwargs["transport"] = httpx2.WSGITransport(app=app_value)
-
         httpx2.Client.__init__(self, **client_kwargs)
 
         _AssertionClient.__init__(
--- a/authlib/integrations/httpx_client/assertion_client.py
+++ b/authlib/integrations/httpx_client/assertion_client.py
@@ -1,6 +1,6 @@
-import httpx
-from httpx import USE_CLIENT_DEFAULT
-from httpx import Response
+import httpx2
+from httpx2 import USE_CLIENT_DEFAULT
+from httpx2 import Response
 
 from authlib.oauth2.rfc7521 import AssertionClient as _AssertionClient
 from authlib.oauth2.rfc7523 import JWTBearerGrant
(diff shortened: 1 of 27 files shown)


azmeuk (approved):
LGTM

authlib/integrations/httpx_client/assertion_client.py
```diff
@@ -1,6 +1,6 @@
-import httpx
-from httpx import USE_CLIENT_DEFAULT
-from httpx import Response
+import httpx2
+from httpx2 import USE_CLIENT_DEFAULT
+from httpx2 import Response
```
lepture: I think you should use try catch to make it compatible. You can show a deprecated messages when httpx is used instead of httpx2.

declension:
This is great, thanks! For planning, any clues on when the next release will be?

Issue #913, opened 2026-07-23
JsonWebKey.import_key_set raises KeyError on an unknown kty, failing the whole JWK Set (RFC 7517 §5)


RalphBragg:
### Summary

`JsonWebKey.import_key_set()` raises `KeyError: 'AKP'` when any key in the set has a `kty` it does not recognise, so the entire key set fails to import. Per RFC 7517 section 5 an unrecognised key should be ignored, not fatal. A single post-quantum key in a published JWKS therefore takes down verification of the classical keys alongside it.

### Reproduction

A JWK Set with a classical RSA key, an EC key, and one post-quantum ML-DSA key (`kty: "AKP"`, RFC 9964):

```python
from authlib.jose import JsonWebKey
JsonWebKey.import_key_set(mixed_jwks)   # RSA + EC + one {"kty":"AKP", ...}
# KeyError: 'AKP'
```

Measured against Authlib 1.7.2. The same set imports fine once the AKP key is removed, so the failure is attributable to the unknown key, not the fixture. Full fixture and probe available on request.

### Why this matters

RFC 7517 section 5: *"Implementations SHOULD ignore JWKs within a JWK Set that use `kty` values that are not understood by them."* ML-DSA (FIPS 204) key publication is beginning across the ecosystem (Node ships it, Go stdlib in 1.27, RFC 9964 defines the `AKP` JWK form). The moment an issuer adds an ML-DSA key to a JWKS that Authlib consumers read, those consumers stop verifying every signature from that issuer, including the classical ones, on a key publication they do not control. In federated deployments (OpenID Connect, open banking) this is an externally triggered outage.

### Suggested fix

In `import_key_set`, when a member's `kty` has no registered key type, skip that member (optionally warn) instead of raising, so the rest of the set still loads. This matches the RFC section 5 SHOULD and does not change behaviour for recognised keys.

### Offer

Happy to share the cross-language reproduction harness (a mixed classical + AKP JWKS with a control set) and to open a PR with the per-key tolerance and a regression test.

PR #914, opened 2026-07-24
Ignore unknown kty in import_key_set (RFC 7517 §5)

RalphBragg:
### What

`JsonWebKey.import_key_set()` now skips a key whose `kty` is present but not recognised, instead of raising `KeyError` and failing the whole set.

### Why

Today `import_key_set` does `[cls.import_key(k) for k in keys]`; the first key with an unregistered `kty` raises `KeyError` (e.g. `KeyError: 'AKP'`) and the entire set fails to import. RFC 7517 §5 says a processor SHOULD ignore key types it does not understand.

This is becoming a live problem as post-quantum keys (ML-DSA, `kty:"AKP"`, RFC 9964) begin appearing in published JWKS next to RSA/EC keys: a consumer stops verifying **every** signature from that issuer, including the classical ones, the moment such a key is published. Relates to #913.

### Change

- `import_key_set` skips only keys whose `kty` is explicitly present and not in `JWK_KEY_CLS`; a recognised-but-malformed key still raises as before.
- Added `test_jwk_import_key_set_ignores_unknown_kty`.

`tests/jose/test_jwk.py` passes (26 passed). I know `authlib.jose` is deprecated in favour of joserfc; happy to also PR joserfc (it has the same behaviour), but since `authlib.jose` still ships and is widely used pre-2.0, this keeps it safe in the meantime.

### DIFF
--- a/authlib/jose/rfc7517/jwk.py
+++ b/authlib/jose/rfc7517/jwk.py
@@ -51,8 +51,16 @@ class JsonWebKey:
         """
         raw = _transform_raw_key(raw)
         if isinstance(raw, dict) and "keys" in raw:
-            keys = raw.get("keys")
-            return KeySet([cls.import_key(k) for k in keys])
+            keys = []
+            for k in raw["keys"]:
+                # RFC 7517, Section 5: ignore a key whose "kty" is not understood
+                # rather than failing the whole set. Matters as post-quantum keys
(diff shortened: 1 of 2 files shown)


azmeuk:
Hi, thank you for your contribution, but unfortunately`authlib.jose` is currently deprecated and will be removed in the upcoming 1.8 version in favor of [joserfc](https://jose.authlib.org/en/). We won't take any contribution on `authlib.jose` now.

PR #934, opened 2026-08-30
Accept httpx URL objects on OAuth client requests

### LINKED ISSUE #931: support `httpx.URL` objects for urls
Hey.

Not super important but for proper integration with `httpx`/`httpx2` it would perhaps be nice if `httpx.URL` objects would be supported instead of "only" plain strings.

Currently this fails at probably various places.

Unfortunately it's probably not very easy to do this... I mean the simplest solution I can think of is doing `str(url)` in all places where a `httpx.URL`, which would be at least all the functions for the integration of `httpx`.

Cheers,
Chris.

gyanu2507:
httpx/httpx2 pass URL instances into get/request/stream. The Authlib wrappers forwarded those objects through as-is and they broke in places that expected a str.

str() them at the request/stream boundary on the async and sync OAuth2 and assertion clients.

Fixes #931

### DIFF
--- a/tests/clients/test_httpx/test_async_oauth2_client.py
+++ b/tests/clients/test_httpx/test_async_oauth2_client.py
@@ -62,15 +62,16 @@ async def test_add_token_get_request(assert_func, token_placement):
 
 
 @pytest.mark.asyncio
-@pytest.mark.parametrize(
-    "assert_func, token_placement",
-    [
-        (assert_token_in_header, "header"),
-        (assert_token_in_body, "body"),
-        (assert_token_in_uri, "uri"),
-    ],
-)
-async def test_add_token_to_streaming_request(assert_func, token_placement):
+async def test_get_accepts_url_object():
+    from httpx2 import URL
+
+    transport = ASGITransport(AsyncMockDispatch({"a": "a"}))
+    async with AsyncOAuth2Client(
+        "foo", token=default_token, transport=transport
+    ) as client:
+        resp = await client.get(URL("https://provider.test"))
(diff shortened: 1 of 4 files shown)


tests/clients/test_httpx/test_async_oauth2_client.py
```diff
@@ -62,24 +62,16 @@ async def test_add_token_get_request(assert_func, token_placement):
...
 @pytest.mark.asyncio
-@pytest.mark.parametrize(
-    "assert_func, token_placement",
-    [
-        (assert_token_in_header, "header"),
-        (assert_token_in_body, "body"),
-        (assert_token_in_uri, "uri"),
-    ],
-)
-async def test_add_token_to_streaming_request(assert_func, token_placement):
```
lepture, line 73: why did you remove `test_add_token_to_streaming_request`? This is a test for the stream method.

gyanu2507: Nothing intentional — I wrote the URL-object test in its place instead of alongside it. My mistake, restored in 7ef220c.

While putting it back I also parametrized the new test over the three token placements, so a `URL` object gets exercised on the body and uri paths and not just the header one.

`tests/clients/test_httpx/` is 72 passed.

gyanu2507: Restored in 7ef220c. It was only gone briefly while cleaning up leftover junk from the URL-object test; `test_add_token_to_streaming_request` is back on the tip.

